PR #122, opened 2026-05-19
fix(oidc): refuse start when default_role would silently grant admin

### LINKED ISSUE #120: oidc.default_role falls back to "admin" when unset — first OIDC login becomes silent admin
In `internal/web/oidc.go:212–215`:

```go
role := o.cfg.DefaultRole
if role == "" {
    role = "admin"
}
```

If an operator enables OIDC without setting `default_role` AND without `allowed_groups` / `allowed_emails`, the very first user who completes an OIDC login is auto-provisioned as **admin**. No warning at startup, no audit indicator. This is surprising even before considering misconfiguration risk — admin is the maximum-privilege role and should require an explicit opt-in.

I'd like your input on the right behavior. Three options I considered:

1. **Fail-fast at startup** when `oidc.enabled: true` AND (`default_role` is unset OR `"admin"`) AND both `allowed_groups` and `allowed_emails` are empty. Operator must explicitly opt in to "anyone-becomes-admin" by setting `default_role: admin` with an explicit empty `allowed_groups: []` and `allowed_emails: []` — a deliberate footprint, not an accident.
2. **Default to `"user"`** when unset. Cleaner-looking but silently changes behavior for any existing deployment that relied on the implicit admin.
3. **Warn at startup**, keep current behavior. Lowest churn, but warnings tend to be ignored.

Option 1 reads least surprising to me and avoids the silent-behavior-change. Happy to draft against whichever you'd prefer.

Cross-reference: `configs/server.example.yml:103` documents the field as `default_role: "admin"`. The example is currently consistent with the implicit fallback; either fix will need that field to become required in the example block too.

ak2k:
## Summary
Closes #120 — refuse `nebula-mgmt` start when `oidc.enabled: true` with `default_role` unset (or `"admin"`) AND no allowlists.

## Implementation
Per your spec on #120:

- `OIDCConfig.Validate()` rejects the unsafe combination with an error that names the three relevant fields so operators see the resolution path inline.
- `LoadServerConfig` calls `Validate()` alongside the existing `tls_cert`/`tls_key` pair check and `ca_auto_rotate` validation.
- `upsertOperator` no longer falls back to `"admin"` when `DefaultRole` is empty — validation guarantees the unset case only reaches here when an allowlist gates it, so `"user"` is the conservative default.
- `configs/server.example.yml`: drops the implicit-fallback hint, sets `default_role: "user"` uncommented, comment explains the rule + the two-field opt-in.

## Test plan
- `OIDCConfig.Validate` covers every line: nil receiver, disabled, both rejected branches (unset + admin), and four accepted branches (user / admin+emails / admin+groups / unset+emails).
- Dedicated test asserts the error message names `default_role`, `allowed_groups`, `allowed_emails`.
- `go test -race -count=1 ./...` clean; `golangci-lint v2.12` 0 issues.

## Breaking change
Existing deployments running implicit-admin will refuse to start. Restore by either setting `default_role: "user"` (recommended) OR `default_role: "admin"` plus at least one allowlist entry.

### DIFF
--- a/configs/server.example.yml
+++ b/configs/server.example.yml
@@ -100,4 +100,10 @@ allow_self_registration: false
 #   groups_claim: "groups"                 # default
 #   allowed_groups: ["nebula-admins"]      # optional allowlist
 #   allowed_emails: []                     # optional allowlist
-#   default_role: "admin"                  # role for newly-federated users
+#   # default_role gates who becomes admin on first OIDC login. The server
+#   # refuses to start if `default_role` is unset (or "admin") AND both
+#   # allowlists are empty — that combination would silently grant admin to
+#   # any successful login. To opt into "anyone-who-can-log-in-is-admin",
+#   # set `default_role: "admin"` AND at least one allowlist entry (the
+#   # two-field footprint is the point).
+#   default_role: "user"                   # role for newly-federated users
--- a/internal/config/oidc_validate_test.go
+++ b/internal/config/oidc_validate_test.go
@@ -1,0 +1,84 @@
+package config
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestOIDCConfig_Validate(t *testing.T) {
+	tests := []struct {
+		name      string
+		cfg       *OIDCConfig
+		wantError string // empty → expect nil; otherwise substring match on err.Error()
+	}{
+		{
+			name: "nil OIDC accepted",
+			cfg:  nil,
+		},
+		{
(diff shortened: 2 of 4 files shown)


juev:
Thanks Adam — well-specced fix that lines up exactly with Option 1 from #120, and the validation test that asserts the error message names `default_role` / `allowed_groups` / `allowed_emails` is a nice touch (operators get the resolution path inline). Merged; release notes for the breaking change will land with the next minor.

PR #124, opened 2026-05-19
chore: tighten data-dir and config.yml permissions (gosec defense-in-depth)

ak2k:
## Summary
`gosec` flags four directory-creation sites at `0o755` and one WriteFile at `0o644` for `config.yml`. Pure defense-in-depth — no advisory associated.

## Changes
- `internal/agent/enroll.go`: `dataDir`, signing-key dir, preflight dir → `0o750`; `config.yml` → `0o640`
- `internal/cli/init.go`: server data dir → `0o750`
- `internal/config/agent.go`: agent config dir → `0o750`

The cert files (`host.crt`, `ca.crt`) stay `0o644` — they're public-by-design (transmitted to peers on handshake). Private keys are already `0o600`. The change narrows the surrounding directories so non-owner local users can't traverse to the cert/key files.

`config.yml` may embed enrollment metadata or refresh state, so treat it as secret-adjacent (`0o640`).

## Test plan
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `go test -race -count=1 ./...` — all 14 packages pass
- [x] `golangci-lint v2.12 run --timeout=5m` — 0 issues
- [x] `gosec ./...` — issue count drops from 32 → 27 (remainder are false positives I'm happy to walk through if useful)

## Notes
Other gosec findings I confirmed as real but NOT addressed here:
- G124 cookie attributes — already disclosed as SA #1 + SA #2
- G710 redirect/URL leak on `operators.go:251` (operator API key in query string) — filed separately as advisory `GHSA-9pg3-25fq-p6cc`

### DIFF
--- a/internal/agent/enroll.go
+++ b/internal/agent/enroll.go
@@ -117,8 +117,9 @@ func Enroll(serverURL, token, dataDir, signingKeyPath string) error {
 		return fmt.Errorf("decode enrollment response: %w", err)
 	}
 
-	// Ensure data directory exists
-	if err := os.MkdirAll(dataDir, 0o755); err != nil {
+	// Ensure data directory exists. 0o750 — agent owns its cert/key files
+	// and no other local user needs to traverse this directory.
+	if err := os.MkdirAll(dataDir, 0o750); err != nil {
 		return fmt.Errorf("create data dir: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


juev:
Thanks Adam — nicely scoped permissions hardening, and I appreciated the explicit call-out that the public cert files stay 0o644 by design. Merged.

Issue #121, opened 2026-05-19
actor.go:40 — legacy config-key auth returns admin=true (separate from #119)


ak2k:
Filing the adjacent observation from #119 as its own issue per your request.

`internal/api/actor.go:34-43`:

```go
func actorIsAdmin(ctx context.Context) bool {
    op := ActorOf(ctx)
    if op == nil {
        return true // legacy config api_key — effectively root
    }
    return op.Role == "admin"
}
```

Combined with `bearerAuth` in `internal/api/middleware.go:69` (which accepts `token == legacyKey` as a fallback and does NOT call `withActor`), any request authenticated via the legacy `cfg.APIKey` config-file fallback gets root access to every endpoint.

Today this is documented behaviour (the legacy single-key auth predates multi-operator support). The question worth deciding independently:

1. Phase the legacy path out entirely — require all production deployments to migrate to operator-bound API keys.
2. Keep it as an emergency-access fallback but with explicit audit on every use ("legacy-admin auth used: $endpoint $method").
3. Gate behind an explicit `legacy_auth.enabled: true` config opt-in so new deployments don't inherit it accidentally.

Not exploitable today as far as I can tell (requires possession of `cfg.APIKey`, which IS the legacy admin secret). Filing as a design discussion per your suggestion.

juev:
Phased out the legacy config-key auth path entirely. Landed in 56c07b7 + da2f3cd on `main`.

## What changed

- `internal/api/middleware.go` — `bearerAuth` no longer accepts a `legacyKey`; the only auth path is now DB-backed `operator_api_keys`. Any request that passes the middleware has a non-nil actor on its context.
- `internal/api/actor.go` — `actorIsAdmin` is fail-closed (`op == nil → false`). `ActorName` returns `"unknown"` instead of `"legacy-admin"` for unauthenticated contexts. The `actor.go:40` root-on-nil branch quoted in this issue is gone.
- `internal/api/server.go` — `NewServer` lost its `apiKey` parameter; `Server` lost the `apiKey` field. All callers updated (`cli/serve.go`, `tests/integration/e2e_test.go`, `internal/api/server_test.go`).
- `internal/api/cas.go` (follow-up cleanup in da2f3cd) — `handleListCAs`, `handleCreateCA`, and `canAccessCA` had defensive `actor == nil` branches that became unreachable once `bearerAuth` started guaranteeing a non-nil actor; those are removed.
[...]

PR #128, opened 2026-05-20
fix(api): audit mobile-bundle authz decisions (closes #119)

### LINKED ISSUE #119: POST /api/v1/hosts/{id}/mobile-bundle accepts any operator key
`internal/api/mobile_bundle.go:19` does no authorization check beyond `bearerAuth`. Any operator API key — including one issued to a low-privilege user managing a different CA's hosts — can mint a fresh certificate + private key for any mobile host on the server. The response inlines the private key (intentional for the mobile-app flow), so there is no follow-up step that could enforce ownership.

Current trust assumption (per ADR 0002 §4.2): "A non-admin operator can … never operate on CAs they do not own." The mobile-bundle endpoint silently bypasses that.

I'd like your input on the right fix before sending a patch. Two shapes that occurred to me:

1. **Admin-or-CA-owner gate**: allow if `actorIsAdmin(ctx)` OR the actor owns the CA that signs the host's network. Reject 403 otherwise; audit-log `host.mobile_bundle.forbidden`.
2. **Admin-only**: strictest, but breaks the model where a non-admin operator manages their own per-CA networks end-to-end.

Option 1 matches the per-operator CA ownership model from ADR 0002 and looks like the right default; happy to draft a PR for that. Let me know which direction you'd prefer.

Adjacent observation, flagging as out of scope here: `internal/api/actor.go:40` returns `actorIsAdmin = true` when the legacy config-key auth path runs (no operator attached). That's a separate concern from this endpoint; mentioning so it's on the radar.

### Why I'm filing as an issue, not an advisory

The endpoint is bearer-auth-gated, so this only affects operators whose threat model includes other operators on the same server. That's the multi-tenant case ADR 0002 introduced — a "feature gap" framing seemed more accurate than "vulnerability".

ak2k:
## Summary

Closes #119. The admin-OR-CA-owner gate on `POST /api/v1/hosts/{id}/mobile-bundle`
(`canAccessHost` → `actorOwnsCA`) was already in place; this PR adds the audit-log
coverage and matrix tests you asked for in the issue comment.

## Changes

`internal/api/metrics.go`: two new audit-action constants in the existing `host.*` group:

```go
auditHostMobileBundleIssued    = "host.mobile_bundle.issued"
auditHostMobileBundleForbidden = "host.mobile_bundle.forbidden"
```

The const-block re-alignment is gofmt's response to the new longest name; struct blocks
below are deliberately left at their prior alignment to keep this PR scoped.

`internal/api/mobile_bundle.go`: wire `recordAuditAction` into both branches:

- 403 path → details `"non_owner"` (matches the bareword-reason style used by `host.auth.failed`)
- 200 path → details `"ca_id=<host.CAID>"` so audit forensics can answer "which CAs minted mobile certs last week" without a join

`internal/api/mobile_bundle_authz_test.go`:

- `TestMobileBundle_RequiresOwnership` (existing) — amended to also assert the `host.mobile_bundle.forbidden` audit entry was written with reason `"non_owner"`
- `TestMobileBundle_AdminCanAccessForeignHost` (new) — exercises the admin override of `canAccessHost`; asserts the issued audit entry with the expected `ca_id`
- `TestMobileBundle_OwningNonAdminCanAccess` (new) — exercises the CA-owner path; asserts the issued audit entry

## Test matrix vs your #119 comment

| Case | Test |
|---|---|
| admin → OK | `TestMobileBundle_AdminCanAccessForeignHost` |
| owning non-admin → OK | `TestMobileBundle_OwningNonAdminCanAccess` |
| non-owner non-admin → 403 | `TestMobileBundle_RequiresOwnership` |
| legacy config-key path → OK | omitted intentionally — that path was removed in #121 (`bearerAuth` is DB-backed only), so there's no scenario left to cover |

## Pre-flight

`make ci` green: vet, build, race tests, golangci-lint v2.12.0 (`0 issues.`).


juev (approved):
Thanks, Adam — the audit-action constants plus the 3-case matrix are exactly the shape I wanted from #119. Appreciate keeping the const-block re-alignment scoped (and leaving the struct blocks alone). Branch is up-to-date with the #129 fix now, CI green, merging.

PR #139, opened 2026-05-21
fix(web): CSRF protection on /ui/* mutating endpoints (GHSA-273q-qgh5-wrj6)

juev:
## Summary

Closes GHSA-273q-qgh5-wrj6 (high severity). Adds stateless double-submit cookie CSRF middleware to all `/ui/*` mutating endpoints. Migrates `/ui/logout` from GET to POST. Rotates token on every privilege transition. Audit-logs rejections as `web.csrf.rejected`.

## Mechanism

- **Token**: 32-byte `crypto/rand`, hex-encoded, carried in non-HttpOnly `_csrf` cookie.
- **Validation**: mutating requests echo the token in `X-CSRF-Token` header (htmx) or `_csrf` form field (HTML forms). Comparison via `crypto/subtle.ConstantTimeCompare`.
- **Rotation**: fresh token issued on `Login`, `StartAuthenticatedSession` (OIDC), `CompleteTwoFactor`, `Logout`.
- **Templates**: `layout.html` adds `<meta name=\"csrf-token\">` + `htmx:configRequest` listener; all forms get `{{ csrfField }}` via per-request FuncMap closure (no handler/data-type changes).
- **Coverage**: pre-auth POST (`/ui/login`, `/ui/register`, `/ui/login/totp`) included — closes login-CSRF.
- **Audit**: `web.csrf.rejected` action with reason (`missing_cookie` / `missing_token` / `mismatch`).

## Known limitation

Documented in `csrf.go` package comment: sibling-subdomain takeover (parent-domain cookie write) can forge a match — inherent to stateless double-submit. Mitigation would require server-side token storage; out of scope for this fix.

## Coordinates with

- **GHSA-rqfj-vv8r-xhqc** (Secure cookie flag): `_csrf` cookie inherits the same `cookieSecure` derivation from `ServerConfig.CookieSecureResolved()`.
- **GHSA-w7w5-5gcp-38rw** (security headers / CSP): existing CSP narrows the XSS surface that could otherwise read `_csrf` cookie (HttpOnly is intentionally false for htmx).

## Files

- New: `internal/web/csrf.go`, `csrf_test.go`, `csrf_rotation_test.go`, `csrf_e2e_test.go`, `logout_test.go`.
- Modified: `internal/web/web.go`, `session.go`, `templates/layout.html` + 13 form templates; ~10 existing handler test files updated with CSRF token plumbing.

## Test plan

- [x] Unit tests in `csrf_test.go` — 13 scenarios (GET bypass, missing cookie/token, header/form match, mismatch, cookie attrs, audit, symmetry, HTML escape).
- [x] Rotation tests in `csrf_rotation_test.go` — 4 privilege transitions.
- [x] E2E tests in `csrf_e2e_test.go` — POST `/ui/cas/{id}/delete` happy + rejected paths (verifies audit entry).
- [x] `go test -race -count=1 ./...` — full suite green.
- [x] `golangci-lint run ./...` — no new findings (existing baseline gosec G120 noise unchanged).
- [x] Independent security-reviewer agent pass — no Critical/Major findings (3 Minor addressed inline).

## Manual verification

Two-tab test (after merging + deploying):
1. Tab A: log in to nebula-mgmt UI.
2. Tab B: open external HTML with `<form action=\"https://.../ui/cas/{id}/delete\" method=POST>` → submit.
3. Expected: `403 Forbidden`, audit log entry `web.csrf.rejected` with `reason=missing_cookie`.
4. Tab A: delete a CA via UI button → succeeds with 303 redirect.

### DIFF
--- a/internal/web/csrf_rotation_test.go
+++ b/internal/web/csrf_rotation_test.go
@@ -77,6 +77,10 @@ func TestCSRF_RotateOnLogin(t *testing.T) {
 	assert.Len(t, newCSRFValue, 64, "CSRF token should be 32 bytes hex-encoded")
 	// Verify rotation: new token should differ from initial
 	assert.NotEqual(t, initialCSRF, newCSRFValue, "Login should rotate CSRF token")
+
+	// Rotation invariant: a form rendered before Login (carrying initialCSRF)
+	// must not validate against the post-rotation cookie.
+	assertPreRotationTokenRejected(t, w, initialCSRF, newCSRFValue)
 }
 
 // TestCSRF_RotateOnOIDCSession verifies that StartAuthenticatedSession
@@ -100,6 +104,20 @@ func TestCSRF_RotateOnOIDCSession(t *testing.T) {
 	}
 	require.NoError(t, err)
 
+	// Establish a pre-rotation CSRF cookie via GET (simulates the browser
+	// having visited the login page before bouncing through OIDC).
+	getReq := httptest.NewRequest(http.MethodGet, "/ui/login", nil)
+	getRec := httptest.NewRecorder()
+	w.ServeHTTP(getRec, getReq)
+	var initialCSRF string
+	for _, cookie := range getRec.Result().Cookies() {
+		if cookie.Name == csrfCookieName {
+			initialCSRF = cookie.Value
+			break
+		}
+	}
+	require.NotEmpty(t, initialCSRF, "GET /ui/login should set CSRF cookie")
+
 	// Simulate OIDC callback by calling StartAuthenticatedSession directly
 	sessionReq := httptest.NewRequest(http.MethodGet, "/ui/callback", nil)
 	sessionRec := httptest.NewRecorder()
@@ -117,6 +135,11 @@ func TestCSRF_RotateOnOIDCSession(t *testing.T) {
 
 	assert.NotEmpty(t, newCSRF, "StartAuthenticatedSession should set CSRF cookie")
 	assert.Len(t, newCSRF, 64, "CSRF token should be 32 bytes hex-encoded")
+	assert.NotEqual(t, initialCSRF, newCSRF, "StartAuthenticatedSession should rotate CSRF token")
+
+	// Rotation invariant: a form rendered before OIDC completion (carrying
+	// initialCSRF) must not validate against the post-rotation cookie.
+	assertPreRotationTokenRejected(t, w, initialCSRF, newCSRF)
 }
 
 // TestCSRF_RotateOnTOTP verifies that CompleteTwoFactor generates a fresh
@@ -185,6 +208,11 @@ func TestCSRF_RotateOnTOTP(t *testing.T) {
 	// Verify rotation: tokens should differ (very likely with random generation)
 	if preTOTPCSRF != "" {
 		assert.NotEqual(t, preTOTPCSRF, postTOTPCSRF, "TOTP completion should rotate CSRF token")
+
+		// Rotation invariant: a form rendered before TOTP completion
+		// (carrying preTOTPCSRF) must not validate against the
+		// post-rotation cookie.
+		assertPreRotationTokenRejected(t, w, preTOTPCSRF, postTOTPCSRF)
 	}
 }
 
@@ -226,4 +254,24 @@ func TestCSRF_ClearOnLogout(t *testing.T) {
 
 	assert.NotNil(t, logoutCSRF, "Logout should clear CSRF cookie")
 	assert.Equal(t, -1, logoutCSRF.MaxAge, "Logout should set CSRF cookie MaxAge=-1")
+}
+
+// assertPreRotationTokenRejected verifies that a form-token issued before
+// rotation no longer validates against the post-rotation cookie. Pins the
+// rotation invariant from both sides: just observing that the new cookie
+// differs from the old is not enough — a regression that emitted a new
+// cookie while still accepting the pre-rotation form-token would pass.
+func assertPreRotationTokenRejected(t *testing.T, w *Web, preRotationToken, postRotationCookie string) {
+	t.Helper()
+	form := url.Values{csrfFormField: {preRotationToken}}
+	req := httptest.NewRequest(http.MethodPost, "/ui/cas/ca-id/delete",
+		strings.NewReader(form.Encode()))
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.AddCookie(&http.Cookie{Name: csrfCookieName, Value: postRotationCookie})
+	rec := httptest.NewRecorder()
+	w.csrfMiddleware(http.HandlerFunc(func(rw http.ResponseWriter, _ *http.Request) {
+		rw.WriteHeader(http.StatusOK)
+	})).ServeHTTP(rec, req)
+	assert.Equal(t, http.StatusForbidden, rec.Code,
+		"pre-rotation form token must not validate against post-rotation cookie")
 }
--- a/internal/web/csrf_rotation_test.go
+++ b/internal/web/csrf_rotation_test.go
@@ -1,0 +1,229 @@
+package web
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/crypto/bcrypt"
+
+	"github.com/juev/nebula-mesh/internal/models"
+)
+
+// TestCSRF_RotateOnLogin verifies that Login generates a fresh CSRF token
+// and sets it in the response.
+func TestCSRF_RotateOnLogin(t *testing.T) {
+	w, store := newTestWeb(t)
+
+	// Create an operator with the test password hash
+	hash, err := bcrypt.GenerateFromPassword([]byte(testPassword), bcrypt.MinCost)
+	require.NoError(t, err)
+	require.NoError(t, store.CreateOperator(context.Background(), &models.Operator{
+		ID:           "test-operator",
+		Username:     "logintest",
+		PasswordHash: string(hash),
+		Status:       models.OperatorStatusActive,
+	}))
+
+	// First, make a GET request to /ui/login to establish initial CSRF cookie
+	getReq := httptest.NewRequest(http.MethodGet, "/ui/login", nil)
+	getRec := httptest.NewRecorder()
+	w.ServeHTTP(getRec, getReq)
+
+	var initialCSRF string
+	for _, cookie := range getRec.Result().Cookies() {
+		if cookie.Name == csrfCookieName {
+			initialCSRF = cookie.Value
+			break
+		}
+	}
+	require.NotEmpty(t, initialCSRF, "GET /ui/login should set CSRF cookie")
+
+	// Now make a POST login request with the initial CSRF token
+	form := url.Values{
+		"username":    {"logintest"},
+		"password":    {testPassword},
+		csrfFormField: {initialCSRF},
+	}
+	loginReq := httptest.NewRequest(http.MethodPost, "/ui/login", strings.NewReader(form.Encode()))
+	loginReq.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	// Add the initial CSRF cookie to the request
+	loginReq.AddCookie(&http.Cookie{
+		Name:  csrfCookieName,
+		Value: initialCSRF,
+	})
+
+	loginRec := httptest.NewRecorder()
+	w.ServeHTTP(loginRec, loginReq)
+
+	// Extract CSRF cookie from login response
+	var newCSRFValue string
+	for _, cookie := range loginRec.Result().Cookies() {
+		if cookie.Name == csrfCookieName {
+			newCSRFValue = cookie.Value
+			break
+		}
+	}
+
+	// Verify that a CSRF token was set in the response
+	assert.NotEmpty(t, newCSRFValue, "Login should set a CSRF cookie")
+	// Verify it's a valid token (non-empty, hex-encoded 32-byte value = 64 chars)
+	assert.Len(t, newCSRFValue, 64, "CSRF token should be 32 bytes hex-encoded")
+	// Verify rotation: new token should differ from initial
+	assert.NotEqual(t, initialCSRF, newCSRFValue, "Login should rotate CSRF token")
+}
+
+// TestCSRF_RotateOnOIDCSession verifies that StartAuthenticatedSession
+// generates a fresh CSRF token.
+func TestCSRF_RotateOnOIDCSession(t *testing.T) {
+	w, store := newTestWeb(t)
+	op := &models.Operator{
+		ID:       "oidc-op",
+		Username: "oidc-user",
+		Status:   models.OperatorStatusActive,
+		// Don't require password_hash for OIDC operators
+	}
+	// For OIDC operators, we need to allow missing password_hash
+	// Actually, let's check what the store requires
+	err := store.CreateOperator(context.Background(), op)
+	if err != nil {
+		// If password_hash is required, we can either set it to empty string
+		// or skip this test. For now, let's set a dummy hash.
+		op.PasswordHash = "$2a$12$abcdefghijklmnopqrstuvwxyzABCDEF"
+		err = store.CreateOperator(context.Background(), op)
+	}
+	require.NoError(t, err)
+
+	// Simulate OIDC callback by calling StartAuthenticatedSession directly
+	sessionReq := httptest.NewRequest(http.MethodGet, "/ui/callback", nil)
+	sessionRec := httptest.NewRecorder()
+	err = w.session.StartAuthenticatedSession(sessionRec, sessionReq, op)
+	require.NoError(t, err)
+
+	// Extract CSRF cookie from OIDC response
+	var newCSRF string
+	for _, cookie := range sessionRec.Result().Cookies() {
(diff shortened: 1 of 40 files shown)


juev:
@ak2k — could I get your review on this one when you have a moment?

This closes the high-severity CSRF advisory (GHSA-273q-qgh5-wrj6) that was disclosed alongside the security headers / OIDC default-role / cross-tenant authz batch you've been involved in. Pattern is stateless double-submit cookie with per-request `Template.Clone()` + FuncMap closure for template integration — no DB changes.

Particularly interested in your take on:

- **Middleware ordering** in `internal/web/web.go` (`csrfMiddleware` sits outside `requireAuth` group to cover pre-auth POSTs; this is intentional for login-CSRF protection, but means invalid CSRF rejections don't bump the rate limiter — design choice, documented in `## Outcomes & Retrospective` of the local plan).
- **Cookie rotation** in `internal/web/session.go` — fresh token on `Login` / `StartAuthenticatedSession` / `CompleteTwoFactor`; cleared on `Logout`. Same `secure bool` derivation as session cookie (inherits the GHSA-rqfj fix).
- **Template FuncMap approach** in the `internal/web/web.go` render helper — `Template.Clone()` per request to scope `csrfToken` / `csrfField` closures to `r.Context()`. Alternatives (data-wrapper, per-handler field) were discussed and rejected; rationale in the squashed commit body.
- **Sibling-subdomain known limitation** — documented in the `internal/web/csrf.go` package comment. Out of scope for this fix.

Independent security-reviewer agent pass returned no Critical/Major; 3 Minor were addressed inline (gosec `#nosec` comments + goimports + `usestdlibvars` cleanups). Full suite green with `-race -count=1`.

Thanks!

internal/web/session.go
```diff
@@ -107,6 +107,15 @@ func (sm *SessionManager) Login(w http.ResponseWriter, r *http.Request, username
 		SameSite: http.SameSiteLaxMode,
 		MaxAge:   cookieMaxAge,
 	})
+
+	// Rotate CSRF token on privilege transition
+	csrfToken, err := generateCSRFToken()
+	if err != nil {
+		slog.Debug("generate CSRF token", "error", err)
```
ak2k: `slog.Debug` is below most production log verbosities. A silently-skipped CSRF rotation on `crypto/rand` error means the browser keeps the pre-privilege-transition token while the session is created and auth succeeds — the rotation invariant the surrounding block guarantees is violated without operator-visible signal. Same pattern at `StartAuthenticatedSession` (line 155) and `CompleteTwoFactor` (line 204). A stricter posture would return the error and fail-closed at the caller, but that changes signatures; this is the minimum.

```suggestion
		slog.Warn("generate CSRF token", "error", err)
```

internal/web/session.go
```diff
@@ -139,6 +148,15 @@ func (sm *SessionManager) StartAuthenticatedSession(w http.ResponseWriter, r *ht
 		SameSite: http.SameSiteLaxMode,
 		MaxAge:   int(sessionDuration.Seconds()),
 	})
+
+	// Rotate CSRF token on privilege transition
+	csrfToken, err := generateCSRFToken()
+	if err != nil {
+		slog.Debug("generate CSRF token", "error", err)
```
ak2k: Same rotation-failure visibility concern as line 114.

```suggestion
		slog.Warn("generate CSRF token", "error", err)
```

internal/web/session.go
```diff
@@ -179,6 +197,15 @@ func (sm *SessionManager) CompleteTwoFactor(w http.ResponseWriter, r *http.Reque
...
```
ak2k: Same rotation-failure visibility concern as line 114.

```suggestion
		slog.Warn("generate CSRF token", "error", err)
```

internal/web/csrf_rotation_test.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+	}
+
+	// Verify that a CSRF token was set in the response
+	assert.NotEmpty(t, newCSRFValue, "Login should set a CSRF cookie")
+	// Verify it's a valid token (non-empty, hex-encoded 32-byte value = 64 chars)
+	assert.Len(t, newCSRFValue, 64, "CSRF token should be 32 bytes hex-encoded")
+	// Verify rotation: new token should differ from initial
+	assert.NotEqual(t, initialCSRF, newCSRFValue, "Login should rotate CSRF token")
+}
```
ak2k, line 84: The rotation invariant the helper protects is that a form rendered *before* rotation (carrying token `initialCSRF` in its `_csrf` field) must fail to validate against the *post-rotation* cookie (`newCSRFValue`). The current test only checks that the new cookie differs from the old, which would still pass under a regression that emits a new cookie and accepts the old one in parallel. Suggested replay assertion below; the same shape should be applied to `TestCSRF_RotateOnOIDCSession` (using `StartAuthenticatedSession`-derived `newCSRF`) and `TestCSRF_RotateOnTOTP` (using `postTOTPCSRF`).

```suggestion
	assert.NotEqual(t, initialCSRF, newCSRFValue, "Login should rotate CSRF token")

	// Rotation invariant: a form rendered before Login (carrying initialCSRF)
	// must not validate against the post-rotation cookie.
	replayForm := url.Values{csrfFormField: {initialCSRF}}
	replayReq := httptest.NewRequest(http.MethodPost, "/ui/cas/ca-id/delete",
		strings.NewReader(replayForm.Encode()))
	replayReq.Header.Set("Content-Type", "application/x-www-form-urlencoded")
	replayReq.AddCookie(&http.Cookie{Name: csrfCookieName, Value: newCSRFValue})
	replayRec := httptest.NewRecorder()
	w.csrfMiddleware(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
		rw.WriteHeader(http.StatusOK)
	})).ServeHTTP(replayRec, replayReq)
	assert.Equal(t, http.StatusForbidden, replayRec.Code,
		"pre-rotation form token must not validate against post-rotation cookie")
}
```

internal/web/csrf_test.go
```diff
@@ -0,0 +1,347 @@
...
+	// Decode to verify it's valid hex
+	decoded, err := hex.DecodeString(token1)
+	require.NoError(t, err)
+	require.Equal(t, 32, len(decoded), "expected 32-byte decoded token")
+
+	// Verify randomness (two calls should differ)
+	token2, err := generateCSRFToken()
+	require.NoError(t, err)
+	assert.NotEqual(t, token1, token2, "expected different tokens on each call")
+}
```
ak2k, line 347: The three pre-auth POST routes (`/ui/login`, `/ui/login/totp`, `/ui/register`) are the specific reason `csrfMiddleware` sits outside `requireAuth` — they need login-CSRF protection. The current battery only exercises `/ui/cas/{id}/delete`. A small table case appended below locks the pre-auth contract through the full router (so middleware ordering is exercised too, not just the helper in isolation).

```suggestion
}

// TestCSRF_PreAuthPOST_RequiresToken verifies that the pre-auth POST routes
// (/ui/login, /ui/login/totp, /ui/register) reject submissions without a
// CSRF cookie+token. These are the routes that justify keeping csrfMiddleware
// outside requireAuth (login-CSRF defense).
func TestCSRF_PreAuthPOST_RequiresToken(t *testing.T) {
	for _, path := range []string{"/ui/login", "/ui/login/totp", "/ui/register"} {
		t.Run(path, func(t *testing.T) {
			w, _ := newTestWeb(t)
			form := url.Values{"username": {"x"}, "password": {"y"}}
			req := httptest.NewRequest(http.MethodPost, path, strings.NewReader(form.Encode()))
			req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
			rec := httptest.NewRecorder()
			w.ServeHTTP(rec, req)
			require.Equal(t, http.StatusForbidden, rec.Code,
				"expected 403 for pre-auth POST %s without CSRF", path)
		})
	}
}
```

ak2k (changes requested):
Walked the diff against `bda27038`, ran the full test suite with `-race` and `golangci-lint run ./...` (both green), and exercised the rotation + e2e tests by hand. Mechanism reads correctly: 32-byte `crypto/rand` token, hex-encoded, constant-time compare, host-only `nebula_csrf` cookie with `Secure` inheriting from the GHSA-rqfj resolver, rotation on every privilege transition, `web.csrf.rejected` audit entries with a useful `reason` field, and the per-request template FuncMap closure cleanly avoids touching every handler's data type. The sibling-subdomain limitation in the package doc (`csrf.go:13-17`) is honest. Pre-auth login-CSRF is covered (`csrfField` present in `templates/login.html`, `login_totp.html`, `register.html`). Multipart bodies safely 403 (`r.ParseForm` ignores multipart, `PostForm.Get("_csrf")` returns `""`, reject path takes the `missing_token` branch). htmx integration via the `layout.html` meta tag + `htmx:configRequest` listener is the right shape.

Four things I'd want addressed before merge, then a handful of lower-priority notes. Suggestion blocks inline below where the file is in the diff; the profile.html one stays here since that file wasn't modified by this PR.

### Would request changes on

1. **`internal/web/templates/profile.html:22` still ships `<a href="/ui/logout" class="btn btn-danger">Sign out</a>`.** The route is POST-only after this PR (`web.go:321`), so clicking that link 405s. Mirror the pattern in `layout.html:243-246`:
   ```html
   <form method="POST" action="/ui/logout" style="display:inline">
       {{ csrfField }}
       <button type="submit" class="btn btn-danger">Sign out</button>
   </form>
   ```

2. **Silent CSRF rotation failure on `crypto/rand` error** at the three rotation sites in `session.go` (`Login`, `StartAuthenticatedSession`, `CompleteTwoFactor`). Each `slog.Debug`s the error and proceeds with the privilege transition — the session is created, auth succeeds, but the browser keeps the pre-transition CSRF cookie. The invariant the rotation hooks exist to guarantee is silently violated. `crypto/rand.Read` failure is rare but not impossible (FIPS, seccomp, exhaustion). Inline suggestions below bump `Debug` → `Warn` on each site; a tighter posture would return the error and let the caller fail-closed, but that's a larger signature change.

3. **The rotation tests never replay the old token after rotation.** `csrf_rotation_test.go` asserts the new cookie differs from the prior cookie, which is necessary but not sufficient. A regression that emitted a new cookie *and* still accepted submissions whose form-token was issued against the pre-rotation cookie would pass. Inline suggestion below adds a post-rotation replay assertion to `TestCSRF_RotateOnLogin`; same pattern applies to `TestCSRF_RotateOnOIDCSession` and `TestCSRF_RotateOnTOTP`.

4. **No direct test for pre-auth POST without CSRF token.** `POST /ui/login`, `/ui/login/totp`, `/ui/register` without a cookie+token should 403 — these are the specific routes that justify keeping `csrfMiddleware` outside `requireAuth`. `csrf_test.go` exercises `/ui/cas/{id}/delete` only; inline suggestion below adds a table case at the end of the file.

### Won't block — flagging for the next pass

- **`csrf.go:82-86` `#nosec G120` justification is incorrect.** `http.Server.MaxHeaderBytes` caps headers, not bodies; there's no `ReadLimit` field on `http.Server`. UI routes have no body cap (only `internal/api/middleware.go:20-29`'s `MaxBytesReader` is wired up). Pre-existing exposure on every UI `ParseForm` site, not introduced here — but the false-claim comment is worth correcting now (and a UI-side `maxBodySize` is a small follow-up).

- **`TestCSRF_RotateOnTOTP` (`csrf_rotation_test.go:124`) bypasses `handleTOTPLogin`.** It flips `op.TOTPEnabled` in memory only and calls `CompleteTwoFactor` directly, so a route-wiring regression in the TOTP flow (csrfMiddleware vs rate-limit interaction, e.g.) wouldn't be caught. A full-route walk through `POST /ui/login` (pending state) → `POST /ui/login/totp` with a valid TOTP would close that gap.

- **Router-mutation ordering in `web.go`.** `WithRateLimiter` rebuilds the router via `setupRoutes`; `WithOIDC` then registers directly on the rebuilt router *outside* the csrf group. Functionally fine (the OIDC callback is GET; `StartAuthenticatedSession` sets the fresh cookie), but the order dependency between `WithOIDC` and `WithRateLimiter` is implicit. Worth documenting on `WithOIDC` so a future reorder doesn't silently drop the routes.

- **E2E coverage is narrow.** `csrf_e2e_test.go` exercises `/ui/cas/{id}/delete` only; the other 20+ mutating routes (host CRUD, network create, settings, operator CRUD, CA retire/rotate, 2FA setup/disable, mobile-bundle generate) go through the same middleware but aren't asserted end-to-end. A small table-driven test (route × method × expectReject/expectAccept) would lock coverage cheaply.

### Nits

- `csrf_rotation_test.go:228` (`TestCSRF_ClearOnLogout`) only asserts `MaxAge=-1`; doesn't check Path/Secure/SameSite symmetry with `setCSRFCookie` — `TestCSRF_SetClearSymmetry` covers it at the helper level, but the integrated logout test could mirror it.
- `TestCSRF_AuditEntryOnReject` (`csrf_test.go:229-250`) explicitly punts on verifying the audit entry; `csrf_e2e_test.go:124+` does the real check. Either promote that pattern up or drop the redundant test.
- `templates/stats_partial.html` is template-cloned every render but references neither `csrfToken` nor `csrfField` — a fast skip in `renderWithStatus` saves ~120 clones/hour/user on the dashboard poll.
- Cookie `Path=/` carries `nebula_csrf` on `/api/*` and `/metrics` requests too; narrowing to `Path=/ui` is hygiene-only.

Happy to send (1)–(4) as a small follow-up PR if you'd prefer one diff over individual suggestion applies. Mechanism overall is solid — these are mostly tightening the testing surface around the rotation invariant.

juev:
Thanks @ak2k — applied all four blocking items.

- **(2)** `Debug` → `Warn` on the three rotation sites: accepted your suggestions directly (`e19d580`, `9f2051f`, `a0040e8`).
- **(4)** Pre-auth POST table-case: accepted your suggestion (`0a48d37`).
- **(1)** profile.html logout `<a>` → POST form with `{{ csrfField }}`, mirroring layout.html. Updated `TestProfilePage_ShowsOperatorDetailsAndLogoutButton` to assert on `action="/ui/logout"` instead of the old `href=` (`06a6edf`).
- **(3)** Extracted your replay assertion into a shared `assertPreRotationTokenRejected` helper and wired it into all three rotation tests. `TestCSRF_RotateOnOIDCSession` now also establishes an initial CSRF via `GET /ui/login` before calling `StartAuthenticatedSession` so the invariant has something to replay against (`06a6edf`).

Full suite green with `-race -count=1`.

On the non-blocking items — I'll land **A** (false `#nosec G120` justification), **C** (`WithOIDC` ordering doc), **F** (drop redundant `TestCSRF_AuditEntryOnReject`), **H** (cookie `Path=/ui`), **E** (`TestCSRF_ClearOnLogout` Path/Secure/SameSite symmetry) as a small follow-up PR right after this one merges.

The larger ones I'll track as separate issues:
- TOTP full-route walk (`POST /ui/login` pending → `POST /ui/login/totp`)
- Table-driven E2E for the remaining ~20 mutating routes
- UI-side `MaxBytesReader` (the real fix behind the corrected `G120` comment)
- `stats_partial.html` clone skip optimization

Ready for another look when you have a moment.

ak2k (approved):
Re-reviewed `06a6edf` against `bda27038`. All four blocking items addressed; here's how each landed.

**(1) profile.html logout → POST form** — `06a6edf` swaps `<a href="/ui/logout">` for `<form method="POST" action="/ui/logout">{{ csrfField }}<button>`. Walked the route table: `/ui/logout` is mounted as `r.Post` inside the requireAuth + csrfMiddleware group (`internal/web/web.go:321`), so the same-session form submission picks up the validated token. `logout_test.go` already covered `GET → 405` and `POST-with-CSRF → 303`, so the form contract is pinned from both sides. `profile_test.go` updated to assert `action="/ui/logout"` rather than the old href.

**(2) slog.Debug → slog.Warn on three rotation sites** — clean (`e19d580`, `9f2051f`, `a0040e8`). Login / StartAuthenticatedSession / CompleteTwoFactor all surface `crypto/rand` failure at Warn now, so a silent rotation skip is operator-visible.

**(3) Replay-assertion helper across the three rotation tests** — `assertPreRotationTokenRejected` directly invokes `csrfMiddleware` with the pre-rotation form token and post-rotation cookie and expects 403. That's the asymmetric invariant the prior tests didn't pin — a regression that emitted a fresh cookie while still accepting the old form token would have slipped past the bare `assert.NotEqual`. Bonus on the OIDC case: the test now seeds `initialCSRF` via a GET to `/ui/login` (the previous version had no pre-rotation cookie at all) and asserts `NotEqual(initialCSRF, newCSRF)` explicitly. Helper sits in the same file as the tests it serves.

**(4) Pre-auth POST table case** — `0a48d37` adds `TestCSRF_PreAuthPOST_RequiresToken` covering `/ui/login`, `/ui/login/totp`, `/ui/register`. Locks the contract that justifies keeping `csrfMiddleware` outside `requireAuth`.

`go test ./... -count=1 -race` and `golangci-lint run ./...` both green on `06a6edf`.

Approving for merge.

— Non-blocking follow-up: the slog.Warn change makes a CSRF-token generation failure observable, but a `generateCSRFToken()` error in any of the three rotation sites still completes login with the pre-rotation cookie intact. Happy to file a separate issue if you'd like a fail-closed variant (clear cookie + 500 / force re-auth on rand.Read failure).

juev:
Thanks a lot for the thorough re-review, @ak2k — really appreciate the deep walk through the rotation invariant and the replay-assertion suggestion. Merging now.

For the non-blocking follow-up (fail-closed on `generateCSRFToken()` rand.Read failure during rotation): yes, please file it as a separate issue — it's a clean scope for a follow-up PR.

PR #140, opened 2026-05-21
fix(audit): validate URL params before audit + redirect in operator handlers

ak2k:
## Summary

Admin handlers across `internal/web/operators.go` and `internal/api/operators.go` reflected URL params (operator `id`, api-key `kid`) into audit `Resource`/`Details` fields without confirming the params matched real entities. Forensic queries on `Resource` saw orphan rows. A related (observability-flavoured) inconsistency in the same UI handler family — `handleOperatorCreateAPIKey` swallowing all `GetOperator` errors as 404 without logging — is bundled in the second commit.

Concrete consequences pre-fix:
- `POST /ui/operators/{garbage-id}/disable` → 303 redirect + audit row with `resource=<garbage-id>`.
- `POST /ui/operators/{A}/api-keys/{kid-of-B}/revoke` → revokes B's real key, audit row says `resource=A details=kid-of-B`. Cross-operator-kid pollution.
- `DELETE /api/v1/operators/{A}/api-keys/{kid-of-B}` → same as above on the JSON API surface (`internal/api/operators.go:174-192` pre-fix).
- `POST /ui/operators/{id}/api-keys` with a DB outage on the operator lookup → 404 "operator not found" instead of 500 "internal error", with nothing logged. Operators chased a phantom missing-data bug instead of a real outage.

The pattern in `handleOperatorResetPassword` (`operators.go:191`) was already correct (GetOperator first, discriminate ErrNotFound from internal-error + log, 404 on NotFound). This PR brings the other four handlers in the file in line, adds a new `Store.GetOperatorAPIKey` to support kid-ownership verification, and covers the matching defect on the JSON API side.

## What changed

- **UI** (`internal/web/operators.go`): `handleOperatorDisable`, `handleOperatorEnable`, `handleOperatorRevokeAPIKey`, `handleOperatorCreateAPIKey` all now `GetOperator(id)` first, discriminate `ErrNotFound` from internal errors (404 vs 500 + log), and proceed only on success. `handleOperatorRevokeAPIKey` additionally calls `GetOperatorAPIKey(kid)` and verifies `key.OperatorID == id` before revoking; early-returns 303 if the kid is already revoked (skips the no-op UPDATE).
- **JSON API** (`internal/api/operators.go`): `handleRevokeOperatorAPIKey` mirrors the UI Revoke fix — `GetOperator(id)` + `GetOperatorAPIKey(kid)` + ownership check + early-return on already-revoked. (API-side `handleDisableOperator` and `handleEnableOperator` were already correct, only Revoke needed patching.)
- **Store interface** (`internal/store/store.go`, `internal/store/sqlite_operators.go`): new `GetOperatorAPIKey(ctx, keyID)` returning the key regardless of revoked state. Only `SQLiteStore` implements `Store`; no mocks to update.

## Behaviour changes visible to clients

| Request | Pre-fix | Post-fix |
|---------|---------|----------|
| `POST /ui/operators/{bogus}/disable` | 303 + bogus audit row | 404, no audit row |
| `POST /ui/operators/{bogus}/enable` | 303 + bogus audit row | 404, no audit row |
| `POST /ui/operators/{A}/api-keys/{kid-of-B}/revoke` | 303 + key revoked + audit row `resource=A` | 404 + key stays + no audit row |
| `POST /ui/operators/{id}/api-keys` under DB outage | 404 "operator not found", nothing logged | 500 "internal error" + slog.Error |
| `DELETE /api/v1/operators/{A}/api-keys/{kid-of-B}` | 204 + key revoked + audit row `resource=A` | 404 + key stays + no audit row |
| `DELETE /api/v1/operators/{id}/api-keys/{already-revoked-kid}` | 404 (ErrNotFound from `WHERE revoked_at IS NULL`) | 204 (idempotent no-op, REST-conventional) |

The last row is a real API client-visible behaviour change. Matches the canonical REST shape for idempotent DELETE.

## Tests

- `internal/web/operator_existence_validation_test.go` (new, 8 tests): Disable/Enable/Revoke/CreateAPIKey-non-existent → 404, cross-operator-kid → 404 (with body wording assertion to seal the existence side channel), already-revoked idempotent (explicit-second path AND disable-cascade path), happy-path audit rows correct (positive assertion locking the `AddAuditEntry` call sites).
- `internal/api/operators_audit_validation_test.go` (new, 4 tests): same scenario battery on the JSON API surface.
- `internal/store/sqlite_operators_test.go`: `TestGetOperatorAPIKey` with three sub-cases pinning the "returns regardless of revoked state" invariant the ownership check relies on.

## Test plan

- [x] `go test ./... -race -count=1` — full suite green (`internal/api` 41s, `internal/web` 50s, all others <10s).
- [x] `golangci-lint run ./...` — 0 issues.
- [x] `gosec ./...` — 0 issues (pre-push hook also runs vet + govulncheck).
- [x] Manual: pre-existing happy-path UI flows (`TestOperators_AdminLifecycle` walks Disable + Revoke on a real operator/key) continues to pass — happy path semantics unchanged.

## Scope decisions

- **Bundled both UI and JSON API surfaces** since the audit-integrity invariant ("audit row reflects the key's real owner") only holds if both surfaces validate. Splitting would have re-opened the bug on whichever surface lagged.
- **Bundled the `handleOperatorCreateAPIKey` error-discrimination fix** (second commit) since it's the same file and the same 8-line pattern. Different theme (observability vs. audit-integrity) but combining saves a second review pass.

### DIFF
--- a/internal/api/operators.go
+++ b/internal/api/operators.go
@@ -210,6 +210,16 @@ func (s *Server) handleRevokeOperatorAPIKey(w http.ResponseWriter, r *http.Reque
 		return
 	}
 	if err := s.store.RevokeOperatorAPIKey(r.Context(), kid); err != nil {
+		if errors.Is(err, store.ErrNotFound) {
+			// Concurrent revoke (another admin, or a DisableOperator
+			// cascade) won the CAS between the GetOperatorAPIKey
+			// snapshot above and this UPDATE. Already revoked — mirror
+			// the RevokedAt != nil idempotent branch above (204, no
+			// audit row, since this actor's call did not cause the
+			// state change).
+			w.WriteHeader(http.StatusNoContent)
+			return
+		}
 		s.logger.Error("revoke api key", "error", err)
 		writeError(w, http.StatusInternalServerError, "failed to revoke api key")
 		return
--- a/internal/api/operators.go
+++ b/internal/api/operators.go
@@ -178,11 +178,38 @@ func (s *Server) handleRevokeOperatorAPIKey(w http.ResponseWriter, r *http.Reque
 	}
 	id := chi.URLParam(r, "id")
 	kid := chi.URLParam(r, "kid")
-	if err := s.store.RevokeOperatorAPIKey(r.Context(), kid); err != nil {
+	if _, err := s.store.GetOperator(r.Context(), id); err != nil {
+		if errors.Is(err, store.ErrNotFound) {
+			writeError(w, http.StatusNotFound, "operator not found")
+			return
+		}
+		s.logger.Error("get operator", "error", err)
+		writeError(w, http.StatusInternalServerError, "failed to look up operator")
+		return
+	}
+	// Verify the API key belongs to this operator before revoking, so the
+	// audit row reflects the key's real owner — not whichever {id} the
+	// caller supplied in the URL.
+	key, err := s.store.GetOperatorAPIKey(r.Context(), kid)
+	if err != nil {
 		if errors.Is(err, store.ErrNotFound) {
 			writeError(w, http.StatusNotFound, "api key not found")
 			return
 		}
+		s.logger.Error("get api key", "error", err)
+		writeError(w, http.StatusInternalServerError, "failed to look up api key")
+		return
+	}
+	if key.OperatorID != id {
+		writeError(w, http.StatusNotFound, "api key not found")
+		return
+	}
+	if key.RevokedAt != nil {
+		// Already revoked. API stays idempotent — 204 with no new audit row.
+		w.WriteHeader(http.StatusNoContent)
+		return
+	}
+	if err := s.store.RevokeOperatorAPIKey(r.Context(), kid); err != nil {
 		s.logger.Error("revoke api key", "error", err)
 		writeError(w, http.StatusInternalServerError, "failed to revoke api key")
 		return
--- a/internal/api/operators_audit_validation_test.go
+++ b/internal/api/operators_audit_validation_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"net/http"
 	"net/http/httptest"
+	"sync"
 	"testing"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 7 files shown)


juev (changes requested):
Nice fix on the audit-integrity invariant — cross-operator-kid pollution was a real forensic-correlation gap, and sealing the body-wording side channel in the test is a good touch.

One thing worth catching before merge: the new `GetOperatorAPIKey` → `RevokeOperatorAPIKey` sequence introduces a small TOCTOU regression vs. the old code path on concurrent revokes.

`RevokeOperatorAPIKey` is an atomic CAS on the SQL side (`UPDATE … WHERE id=? AND revoked_at IS NULL`) and returns `ErrNotFound` when 0 rows are affected. If another admin — or a `DisableOperator` cascade — revokes the same kid between our `GetOperatorAPIKey` (where `RevokedAt == nil`) and our `RevokeOperatorAPIKey`, the second call now falls into the default 500 branch instead of being treated as idempotent.

The old UI handler had a `case errors.Is(err, store.ErrNotFound): // UI stays idempotent` arm that absorbed exactly this race. The new code lost it on both surfaces.

Failure mode is safe (key is revoked, first actor's audit row is written), but the user-visible result flips from `303`/`204` to `500` in the race window. Two-line fix on each surface restores the old graceful behavior:

**`internal/web/operators.go`** (handleOperatorRevokeAPIKey):
```go
if err := w.store.RevokeOperatorAPIKey(r.Context(), kid); err != nil && !errors.Is(err, store.ErrNotFound) {
    w.logger.Error("revoke api key", "error", err)
    http.Error(rw, "internal error", http.StatusInternalServerError)
    return
}
```

**`internal/api/operators.go`** (handleRevokeOperatorAPIKey):
```go
if err := s.store.RevokeOperatorAPIKey(r.Context(), kid); err != nil && !errors.Is(err, store.ErrNotFound) {
    s.logger.Error("revoke api key", "error", err)
    writeError(w, http.StatusInternalServerError, "failed to revoke api key")
    return
}
```

Race window is narrow and probably never hit in practice — happy to merge as-is and follow up, but figured I'd flag it since the old code handled it explicitly.

ak2k:
Pushed the TOCTOU fix as `485dcc6`. Both surfaces (UI `handleOperatorRevokeAPIKey`, API `handleRevokeOperatorAPIKey`) now route `store.ErrNotFound` from the `RevokeOperatorAPIKey` CAS into the same idempotent response as the `RevokedAt != nil` early-return branch above — 303 on the UI, 204 on the API.

One deviation from your literal suggestion worth flagging: I split the `errors.Is(err, store.ErrNotFound)` case into an explicit branch rather than using the inline `err != nil && !errors.Is(...)` pattern, because the latter falls through to write an audit row attributing the revoke to the race-loser. Mirroring the no-audit-row behavior of the existing `RevokedAt != nil` branch felt like the right call for forensic correlation — but if you'd rather match the literal pattern, that's a 5-line revert and I'm fine either way.

Regression coverage uses a small `concurrentRevokeStore` wrapper that revokes the row directly between `GetOperatorAPIKey` and `RevokeOperatorAPIKey` to exercise the race deterministically (no goroutines / timing). New tests:
- `TestAPI_HandleRevokeOperatorAPIKey_ConcurrentRevokeRace_Idempotent`
- `TestHandleOperatorRevokeAPIKey_ConcurrentRevokeRace_Idempotent`

`go test ./... -count=1 -race` and `golangci-lint run ./...` both green.

juev (approved):
Verified `485dcc6` against the prior review note. CAS-race branch lands cleanly on both surfaces (`internal/web/operators.go:323-332`, `internal/api/operators.go:213-222`): `ErrNotFound` from the `RevokeOperatorAPIKey` CAS now mirrors the `RevokedAt != nil` early-return — 303/204, no audit row.

Good call on deviating from the literal `err != nil && !errors.Is(...)` pattern I sketched. That version falls through to `AddAuditEntry` and writes a row attributing the revoke to the race-loser, which would mislead forensic correlation. Your explicit `ErrNotFound` branch skips audit entirely, which is the correct semantics: only the actor whose UPDATE actually flipped the row gets credit.

The `concurrentRevokeStore` wrapper is a clean way to pin the TOCTOU window deterministically — `raceOnce` guarantees exactly one CAS injection between snapshot and UPDATE, the embedded `store.Store` keeps the rest of the surface intact, and the assertion that the race-loser writes zero audit rows directly locks the invariant. Same shape works for both UI and API surfaces; nice symmetry.

On the non-blocking follow-up from #139 (fail-closed on `generateCSRFToken()` rand.Read failure during rotation) — yes please, separate issue is the right scope.

LGTM.

PR #147, opened 2026-05-23
fix(api): re-check operator status in admin authz gates

ak2k:
## Summary

`bearerAuth` (`internal/api/middleware.go:35`) loads the operator on request entry and attaches them to `r.Context()` via `withActor`. Downstream admin authz gates use `actorIsAdmin(ctx)`, which reads that captured snapshot for the duration of the request. If another admin runs `DisableOperator` (`internal/store/sqlite_operators.go:257` — atomic txn: status flip + sessions deleted + API keys revoked) on the captured operator between bearerAuth and the handler's mutating SQL, the in-flight handler still sees them as an active admin and the mutation proceeds.

The same gap covers the `actorOwnsCA` short-circuit in `internal/api/authz.go`, which feeds `canAccessCA` / `canAccessHost` / `canAccessNetwork`, so destructive handlers like `handleDeleteCA` / `handleRotateCA` / `handleDeleteHost` inherit the stale-admin bypass.

## Approach

I added `(*Server).isActiveAdmin(ctx)` — re-fetches the captured operator via `s.store.GetOperator` and returns true only if the fresh row is active+admin. Fails closed on `ErrNotFound` or transient errors.

I then swapped all 12 admin authz gates uniformly from `actorIsAdmin(r.Context())` to `s.isActiveAdmin(r.Context())`. The bare `actorIsAdmin` function is preserved as a building block.

| file | sites |
|---|---|
| `operators.go` | handleCreateOperator, handleListOperators, handleDisableOperator, handleEnableOperator, handleCreateOperatorAPIKey, handleRevokeOperatorAPIKey, handleListOperatorAPIKeys |
| `networks.go` | handleCreateNetwork, handleListNetworks scoping |
| `hosts.go` | handleListHosts scoping, handleGetBlocklist |
| `cas.go` | handleListCAs, canAccessCA |
| `settings.go` | handleGetSettings, handlePatchSettings |
| `audit.go` | handleGetAuditLog |
| `authz.go` | actorOwnsCA short-circuit |

## What this does not fix

The TOCTOU between the re-check and the handler's subsequent mutating SQL is narrowed but not closed — a disable that commits in that window still permits the mutation. Closing it would need `BEGIN IMMEDIATE` wrapping check+mutation or `WHERE status=Active` predicates on each mutating statement. Both are larger refactors I'd prefer to scope separately; the re-check covers the bulk of the exposure window.

I also looked at role-mutation paths (no `UpdateOperatorRole` in the store, no PATCH handler, OIDC `upsertOperator` returns existing operators unchanged on re-login). `isActiveAdmin` already re-reads `.Role` from the fresh row, so when a role-mutation surface lands, the helper covers it without further changes.

## Cost

Each gate now does one indexed PK lookup against `operators`. For request paths that hit a single authz gate (the common case), this is one extra round-trip on top of bearerAuth's existing API-key lookup. No handler currently calls `isActiveAdmin` more than once per request, so a request-lifetime cache would save zero round-trips — and would also weaken the disabled-mid-flight detection that is the entire point. I skipped the cache.

## Test plan

- [x] `go test ./... -race -count=1` — all packages green
- [x] `golangci-lint run ./...` (v2.12.2, CI-pinned) — 0 issues
- [x] `internal/api/inflight_authz_test.go` pins the disabled and deleted-row branches of the gate against a stale captured ctx
- [x] `internal/web/oidc_disable_cascade_test.go` is a provider-explicit regression pin for the OIDC session cascade on disable (the local-auth cascade is already covered in `internal/store/sqlite_operators_test.go::TestDisableOperator_CascadesSessionsAndAPIKeys`)
- [x] `internal/api/authz_test.go::TestActorOwnsCA_Admin` updated to seed the in-memory admin into the store (previously relied on the snapshot-only check; the seed makes the test work under the new re-fetch path)

Note: `make sec-sweep` reports 2 pre-existing G124 findings in `internal/web/csrf.go:138/150` from #139. Not touched by this PR; raising separately.

### DIFF
--- a/internal/api/audit.go
+++ b/internal/api/audit.go
@@ -10,7 +10,7 @@ import (
 const defaultAuditLimit = 100
 
 func (s *Server) handleGetAuditLog(w http.ResponseWriter, r *http.Request) {
-	if !actorIsAdmin(r.Context()) {
+	if !s.isActiveAdmin(r.Context()) {
 		writeError(w, http.StatusForbidden, "audit log access requires the admin role")
 		return
 	}
--- a/internal/api/authz.go
+++ b/internal/api/authz.go
@@ -8,10 +8,29 @@ import (
 	"github.com/juev/nebula-mesh/internal/store"
 )
 
+// isActiveAdmin re-fetches the captured-ctx actor and reports whether
+// they are still an active admin. Round-trips the operators table per
+// call. Closes the stale-snapshot gap but not the residual TOCTOU
+// between this check and the handler's subsequent mutating SQL.
+func (s *Server) isActiveAdmin(ctx context.Context) bool {
+	captured := ActorOf(ctx)
+	if captured == nil {
+		return false
+	}
+	fresh, err := s.store.GetOperator(ctx, captured.ID)
+	if err != nil {
+		if !errors.Is(err, store.ErrNotFound) {
+			s.logger.Error("isActiveAdmin: store lookup", "operator", captured.ID, "error", err)
+		}
+		return false
+	}
+	return fresh.Status == models.OperatorStatusActive && fresh.Role == "admin"
+}
+
 // actorOwnsCA returns true if the actor in ctx is admin, or owns the CA with caID.
 // Returns (false, nil) for empty caID or ErrNotFound. Errors only for unexpected DB errors.
 func (s *Server) actorOwnsCA(ctx context.Context, caID string) (bool, error) {
-	if actorIsAdmin(ctx) {
+	if s.isActiveAdmin(ctx) {
 		return true, nil
 	}
 	if caID == "" {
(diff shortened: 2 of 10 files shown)


juev (approved):
LGTM — all 12 admin authz gates swapped uniformly, isActiveAdmin fails closed on every error class, both disabled-mid-flight and deleted-operator branches pinned in the new inflight test.

juev:
Thanks Adam — really thorough sweep here. All 12 admin authz gates swapped uniformly, with `isActiveAdmin` failing closed on every error class (not just `ErrNotFound`), which is exactly the right posture for a re-check primitive. The `inflight_authz_test.go` pinning both the disabled-mid-flight and deleted-operator branches through the actual handler path (not just the helper) is the right shape of test for this kind of fix.

Also appreciate the explicit scope note on the remaining TOCTOU window between the re-check and the mutating SQL, and the rationale for skipping the request-lifetime cache (it would weaken the detection the gate exists for). Those scope decisions — and the OIDC cascade regression pin — are what makes this a tidy PR rather than a sprawling one. Great work!

PR #149, opened 2026-05-24
fix(store): enforce network-scoped overlay-IP uniqueness (migration 018)

ak2k:
## Problem

Migration 014 moved host addresses into a `host_addresses` table and dropped the `UNIQUE(network_id, nebula_ip)` constraint migration 001 had. The only remaining guard is `validateHostIPs`, an API-layer read-then-check — a TOCTOU window that two concurrent host creates can both pass, leaving two hosts bound to the same overlay IP. In a Nebula mesh that's a security defect, not just a data-integrity one: two hosts sharing an overlay address means ambiguous routing and a CA that has issued certs letting one host receive another's traffic.

## Fix

- **Migration 018** restores the invariant declaratively: denormalize `network_id` onto `host_addresses` (immutable post-create) + a `UNIQUE(network_id, address)` index.
- `setHostAddresses` inserts with `ON CONFLICT(network_id, address) DO NOTHING` and maps a zero-row result to `ErrIPTaken` — *targeted*, so `PRIMARY KEY` / `NOT NULL` violations still surface rather than being swallowed by a blunt `INSERT OR IGNORE`. The create/update handlers map `ErrIPTaken` → **409**.
- The migration is intentionally **fail-loud** on a pre-existing duplicate IP — it refuses to start until an operator resolves the defect (there's no safe automatic "which host wins") — and the statement-by-statement loader self-heals on re-run.

## Tests

Serial reject, concurrent exactly-one-wins (`-race`, file-backed so the connection pool actually grows), same-IP-different-networks allowed, plus a direct migration test covering backfill of pre-existing rows, the fail-loud-on-duplicate path, and the down round-trip.

---

Part of the #8c race-hardening work, split into two independent PRs; this is the host-IP-uniqueness half (the enrollment-token-atomicity half is a separate PR).

### DIFF
--- a/internal/api/hosts.go
+++ b/internal/api/hosts.go
@@ -160,6 +160,14 @@ func (s *Server) handleCreateHost(w http.ResponseWriter, r *http.Request) {
 		CreatedAt: now,
 	}
 	if err := s.store.CreateHostAndToken(r.Context(), host, token); err != nil {
+		// Reachable only via a TOCTOU race between concurrent host writes: the
+		// validateHostIPs fast-path above returns 400 for any collision visible
+		// at request time, so reaching the store's UNIQUE(network_id, address)
+		// guard means another writer claimed the IP in between.
+		if errors.Is(err, store.ErrIPTaken) {
+			writeError(w, http.StatusConflict, "one or more nebula_ips are already assigned to another host in this network")
+			return
+		}
 		s.logger.Error("create host and token", "error", err)
 		writeError(w, http.StatusInternalServerError, "failed to create host")
 		return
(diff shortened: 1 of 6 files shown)


juev (approved):
Thanks for restoring the network-scoped overlay-IP uniqueness invariant declaratively via migration 018 — the targeted `ON CONFLICT(network_id, address) DO NOTHING` → `ErrIPTaken` mapping (so PK/NOT NULL still surface) and the concurrent exactly-one-wins test are solid. FYI for whoever deploys: the migration is intentionally fail-loud on a pre-existing duplicate IP, so a DB that already accumulated one via the old TOCTOU window will refuse to start until an operator resolves it — correct behavior, just worth flagging in release notes. LGTM.

PR #150, opened 2026-05-24
fix(api): consume enrollment token atomically with host enroll

ak2k:
## Problem

`handleEnroll` consumed the single-use enrollment token up front in its own committed transaction, *before* the CA signature and certificate save. Any transient failure in those later steps (CA error, DB blip) permanently **burned** the token — the host could never re-enroll without an operator minting a fresh one.

## Fix

- New read-only `GetEnrollmentToken` peek resolves + validates the token (same `ErrNotFound` / `ErrTokenUsed` / `ErrTokenExpired` sentinels) *before* signing.
- New `ConsumeTokenAndEnrollHost` folds the CAS consume (`UPDATE … SET used=1 WHERE used=0`) + cert save + host enroll into **one transaction** — the token is marked used *iff* enrollment persists.
- `handleEnroll` is reordered to consume-last (peek → sign → atomic consume+enroll). A concurrent enrollment racing the same token loses the CAS → `ErrTokenUsed` → **409**.
- A shared `enrollHostInTx` keeps the token-less mobile-bundle path (`SaveCertificateAndEnrollHost`) unchanged.

Net: a transient failure now leaves the token reusable for a clean retry, and single-use is preserved (strengthened, even, under concurrency).

## Tests

Store-level peek-doesn't-consume, no-burn-on-enroll-failure, concurrent-same-token exactly-one-enrolls (`-race`, file-backed), expired peek; plus an API-level replayed-token → 409.

---

Part of the #8c race-hardening work, split into two independent PRs; this is the enrollment-token-atomicity half (the host-IP-uniqueness half is a separate PR). Note: the existing post-enroll `UpdateHostSigningPub` step remains outside the enroll transaction — pre-existing behavior, not widened by this change.

### DIFF
--- a/internal/api/enroll.go
+++ b/internal/api/enroll.go
@@ -45,8 +45,11 @@ func (s *Server) handleEnroll(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Consume token (validates one-time use and expiry)
-	tok, err := s.store.ConsumeToken(r.Context(), req.Token)
+	// Resolve + validate the token (one-time use, expiry) WITHOUT consuming it
+	// yet. The single-use consume happens atomically with the cert save in
+	// ConsumeTokenAndEnrollHost below, so a failure mid-enrollment does not burn
+	// the token (#8c enrollment-token rollback atomicity).
+	tok, err := s.store.GetEnrollmentToken(r.Context(), req.Token)
 	if errors.Is(err, store.ErrNotFound) {
 		s.metrics.recordEnrollment(resultDenied)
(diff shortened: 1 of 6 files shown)


juev (approved):
Thanks — deferring the token consume to a peek → sign → atomic consume+enroll flow fixes the token-burn-on-transient-failure cleanly, and folding the CAS (`UPDATE … WHERE used=0`) + cert save + enroll into one transaction preserves single-use even under concurrent enroll races. Nice that the shared `enrollHostInTx` keeps the mobile-bundle path unchanged. LGTM.

PR #151, opened 2026-05-25
test(api): pin multi-tenant read-side scoping (property harness + boundary battery)

ak2k:
## What

A property-test harness that pins the read-side multi-tenant scoping surface, plus the boundary battery. Test-only — no production code changes.

The keystone is a registry classifying every registered `GET` route, paired with a `chi.Walk` test that fails the moment a new GET route appears unclassified. The cross-operator data leak fixed in GHSA-598g-h2vc-h5vg slipped in because new read paths were not forced to declare their scoping; this turns that omission into a build failure rather than an advisory.

The same registry drives the assertions:

- **collection reads** (`/hosts`, `/networks`, `/cas`) — a non-admin sees only resources under CAs they own; an admin sees all; neither operator sees the other's data or the admin-owned default CA.
- **admin-only reads** (`/operators`, `/operators/{id}/api-keys`, `/audit-log`, `/blocklist`, `/settings`) — non-admins get 403 with no tenant data in the body.

Boundary battery:

- `?network_id=<foreign>` cannot be used to read another operator's hosts (the store filter runs before the owned-CA scoping).
- `/operators` 403 does not enumerate operators.
- a non-owner never receives a single-resource body — host, network, firewall, or CA.

## Notes

`go test ./... -race` and `golangci-lint run ./...` both green.

While writing the by-id tests I noticed a low-severity existence side-channel: a foreign-but-existing host returns 403 while a missing id returns 404, so the status distinguishes existence. It is systemic across the by-id host handlers. The test asserts the non-leak invariant rather than pinning the 403/404 split, so a future uniform-404 hardening will not have to fight it — happy to follow up separately if you would like that change.

### DIFF
--- a/internal/api/scoping_boundary_test.go
+++ b/internal/api/scoping_boundary_test.go
@@ -1,0 +1,112 @@
+package api
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestListHosts_ForeignNetworkFilterDoesNotBypassScoping pins that a
+// non-admin cannot exfiltrate another operator's hosts by passing the
+// foreign network's id as the ?network_id= filter. The filter runs in the
+// store query, but the owned-CA scoping in handleListHosts runs afterwards
+// (hosts.go:190), so the result must stay empty for B even when the filter
+// matches A's hosts.
+func TestListHosts_ForeignNetworkFilterDoesNotBypassScoping(t *testing.T) {
+	srv, st := newTestServer(t)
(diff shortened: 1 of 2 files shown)


juev (approved):
Thanks — turning the "new GET route must declare its scoping" omission into a build failure via the `chi.Walk` registry is a great guard against another GHSA-598g-style read-path leak. The collection/admin-only/by-id boundary battery is thorough, and deliberately asserting the non-leak invariant instead of pinning the 403-vs-404 split is the right call so a future uniform-404 hardening won't fight it. LGTM.

PR #152, opened 2026-05-25
test(pki): cover CA key decryption, destruction, and signer boundaries

ak2k:
## What

Test-only coverage for the CA key lifecycle — no production code changes.

`internal/pki/resolver.go` had no test file, yet `LoadByID` is the on-demand key-decryption path: it unwraps the DEK with the master key, opens the AEAD-sealed signing key, and returns a live `CAManager`. This adds:

- **round-trip**: the decrypted private key matches the stored CA cert's public key (proves the real key was recovered, not garbage that merely parsed)
- **destruction on the decryption path**: a `LoadByID` manager, once `Wipe()`d, can no longer mint a cert that verifies — the GHSA-8h84-fhqq-q58v contract pinned where it matters most
- **no key sharing across CAs**: distinct CAs decrypt to distinct keys, and wiping one manager does not disturb another (the resolver's documented no-cache contract)
- **failure surfaces as an error**: unknown CA, nil receiver/master, and tampered DEK / key-material (AEAD integrity) all error rather than returning a usable-but-wrong CA

Plus signer property pins:

- a cert minted after `Wipe()` fails verification (the `NewCA` construction path)
- minted host certs are never CA certs (`IsCA=false`)
- `notAfter` is clamped to the CA's expiry when a request exceeds it, and left untouched when it doesn't

## Notes

`go test ./... -race` and `golangci-lint run ./...` both green.

### DIFF
--- a/internal/pki/resolver_test.go
+++ b/internal/pki/resolver_test.go
@@ -1,0 +1,265 @@
+package pki_test
+
+import (
+	"bytes"
+	"context"
+	"crypto/ed25519"
+	"crypto/rand"
+	"io"
+	"log/slog"
+	"net/netip"
+	"testing"
+	"time"
+
+	"github.com/slackhq/nebula/cert"
+	"golang.org/x/crypto/curve25519"
+
+	"github.com/juev/nebula-mesh/internal/keystore"
+	"github.com/juev/nebula-mesh/internal/models"
+	"github.com/juev/nebula-mesh/internal/pki"
+	"github.com/juev/nebula-mesh/internal/store"
+)
+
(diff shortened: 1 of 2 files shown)


juev (approved):
Thanks for this — the CA key-lifecycle coverage is exactly where it was missing. The round-trip (decrypted key matches the stored cert's public key) plus the post-`Wipe()` non-verifiable-cert assertions pin the GHSA-8h84 destruction contract on the on-demand decryption path, and the tampered-DEK / no-key-sharing cases lock down the resolver's no-cache guarantee. LGTM.

PR #153, opened 2026-05-25
fix(store): DeleteCA refuses while any ca_id-carrying table references the CA

ak2k:
## Problem

`ca_id` is a plain `TEXT` column with no DB-level foreign key on `networks`, `hosts`, `certificates`, and `blocklist` — it defaults to the empty string for pre-multi-CA rows, so a real FK isn't possible and referential integrity is enforced in application code. But `DeleteCA` only checked `networks`.

Deleting a CA still referenced by a host, certificate, or blocklist row orphaned that row; its `ca_id` then failed to resolve in `caForHost`, surfacing as a 500 on the affected host's next enroll / poll / mobile-bundle. The realistic independent vectors are a blocklist row whose host was deleted (`ON DELETE SET NULL` keeps the row while `ca_id` lingers) and a host whose `ca_id` diverged from its network's via the default-CA fallback.

## Fix

Extend the reference check to all four `ca_id`-carrying tables. The CA is only deletable once nothing references it, and the refusal now names each blocking table. The single-network refusal message is byte-identical to before.

## Tests

`internal/store/sqlite_cas_test.go`: refuses for each of networks / hosts / certificates / blocklist independently (each case isolates its table), reports all blockers when several reference the CA, ignores legacy empty-`ca_id` rows, and succeeds when unreferenced. The host/certificate/blocklist cases fail without the fix.

## Follow-up (out of scope)

The reference checks and the `DELETE` are separate statements (as before this change), leaving a narrow admin-only TOCTOU window; closing it fully would wrap them in a single transaction.

### DIFF
--- a/internal/store/sqlite_cas.go
+++ b/internal/store/sqlite_cas.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"strings"
 	"time"
 
 	"github.com/forgekeep/nebula-mesh/internal/models"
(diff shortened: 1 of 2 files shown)


juev:
Thanks for this — the fix itself is correct and well-tested: extending the reference check to all four `ca_id`-carrying tables (networks/hosts/certificates/blocklist), preserving the byte-identical single-network message, and ignoring legacy empty-`ca_id` rows. The per-table isolation tests and the multi-blocker case are exactly right.

**One thing to fix before we can merge** — a test-helper collision with #149 (now merged). Both declare a package-level `mustExec` in package `store` with different signatures:

- `internal/store/migration_018_test.go` (from #149, on `main`): `func mustExec(t *testing.T, db *sql.DB, q string, args ...any)`
- `internal/store/sqlite_cas_test.go` (this PR): `func mustExec(t *testing.T, s *SQLiteStore, query string, args ...any)`

Each PR was green against `main` on its own, so neither CI surfaced it — but now that #149 has landed, rebasing this PR onto `main` will fail the `store` test build with `mustExec redeclared`.

Could you rebase onto the latest `main` and rename this PR's helper (e.g. `mustExecStore`, since it takes `*SQLiteStore` and has the narrower type)? Alternatively, drop the local helper and reuse the one now on `main` — though the signatures differ, so a rename is the smaller change. Once that's done and CI is green we'll merge. Thanks!

PR #154, opened 2026-05-25
fix(api): scope ListHosts to owned CAs in SQL so the row cap can't undercount

ak2k:
## Problem (work-yzu3)

`handleListHosts` set `HostFilter.Limit = 1000`, called `ListHosts` (which applies `ORDER BY name LIMIT 1000` across **all** CAs), then filtered to the operator's owned CAs **in Go**. Once a deployment has more than 1000 hosts, a non-admin whose own hosts sort after the first 1000 silently sees fewer than they own — the ownership filter runs after the window has already dropped them.

## Fix

- `HostFilter` gains a `CAIDs []string` field. When set, `ListHosts` scopes with `AND ca_id IN (SELECT value FROM json_each(?))` **before** `ORDER BY name / LIMIT`. The CA ids are bound as a single JSON-array parameter, so the query string stays constant — fully parameterized, no dynamic SQL string-building.
- `handleListHosts` computes the non-admin's owned CA ids and passes them as `filter.CAIDs`, so the limit applies after scoping. It short-circuits to `[]` when the actor owns no CAs. The Go post-filter is removed.

Admins are unchanged (no `CAIDs` → unscoped, still capped at 1000 — the global pagination cap is a separate concern). All other `ListHosts` callers pass no `CAIDs`, so their queries are byte-identical.

## Tests

- store: `ListHosts` with `CAIDs` and `Limit: 1` returns the owned host even when a foreign host sorts first (scope-before-limit).
- api: with 1001 foreign hosts sorting ahead of the owner's one host, a non-admin still sees their host via `GET /api/v1/hosts` — the end-to-end work-yzu3 reproduction.

### DIFF
--- a/internal/api/hosts.go
+++ b/internal/api/hosts.go
@@ -180,14 +180,10 @@ func (s *Server) handleListHosts(w http.ResponseWriter, r *http.Request) {
 		Limit:     1000,
 	}
 
-	hosts, err := s.store.ListHosts(r.Context(), filter)
-	if err != nil {
-		s.logger.Error("list hosts", "error", err)
-		writeError(w, http.StatusInternalServerError, "failed to list hosts")
-		return
-	}
-
-	// For non-admin, scope to owned CAs
+	// Non-admins are scoped to hosts under CAs they own. Push that scope into
+	// the query rather than filtering the result in Go: a global LIMIT applied
+	// before the ownership filter could drop an operator's own hosts out of the
+	// window, silently undercounting once the deployment exceeds the limit.
 	if !s.isActiveAdmin(r.Context()) {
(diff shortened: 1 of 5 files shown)


juev (approved):
Thanks — pushing the owned-CA scope into SQL before `ORDER BY name / LIMIT` is the correct fix for the work-yzu3 undercount, and binding the CA ids as a single JSON-array param through `json_each` keeps the query string constant with no dynamic SQL. Verified the nil/empty distinction is airtight: the handler short-circuits to `[]` when the actor owns no CAs, so an empty-non-nil filter never reaches the unscoped path. LGTM.

PR #155, opened 2026-05-25
fix(configgen): round-trippable YAML for non-literal-safe inline PEM (GHSA-7hp6)

ak2k:
## What
`configgen.Generate` could emit a YAML literal-block scalar that neither
`gopkg.in/yaml.v3` nor Nebula's own config loader can parse back, for
inline-PEM content that isn't literal-block-safe (a leading tab/space or
control bytes). `literalString.MarshalYAML` hard-coded `yaml.LiteralStyle`,
which bypasses the encoder's own safety fallback.

This is a residual round-trip gap in the GHSA-7hp6 hardening: that change moved
generation to a typed struct so structural-break characters can't escape, but
forcing literal style on arbitrary content can still produce YAML that won't
reload. Reachability is low (the inline-PEM path is fed server-minted PEMs, not
raw operator input), so it's a latent correctness gap, not a live exploit — but
it contradicts the round-trip guarantee the generator is meant to provide.

## Fix
Gate literal-block style behind a conservative `literalBlockSafe` predicate;
fall back to double-quoted (which always round-trips) otherwise. Well-formed PEM
blocks stay literal-safe and keep their readable multi-line form, so normal
output is unchanged.

## How it was found
`FuzzGenerate` feeds randomized `GeneratorInput` through `Generate`, re-loads
the output through Nebula's real `config.C` parser (the path an agent runs on
reload), and round-trips the scalar fields an agent reads (`am_lighthouse`,
`punchy.punch`, `listen.port`). It surfaced the tab/whitespace case in ~25s.

## Test plan
- `go test ./internal/configgen/ -race` (existing + new regression + predicate tests)
- `go test ./internal/configgen/ -run '^$' -fuzz=FuzzGenerate` — 60s+, clean (407k execs)
- `golangci-lint run ./internal/configgen/`

The fuzz test imports `slackhq/nebula/config`, which adds a few `// indirect`
entries to go.mod (logrus, mergo, yaml v2/v3 — already in the module graph via
Nebula).

## Note
First piece of a broader testing investment (correctness/concurrency/scale/fuzz).
I'll send a short proposal separately and would value your take before the
heavier parts.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 )
 
 require (
+	dario.cat/mergo v1.0.2 // indirect
 	filippo.io/bigmod v0.1.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
@@ -34,7 +35,9 @@ require (
 	github.com/prometheus/common v0.66.1 // indirect
 	github.com/prometheus/procfs v0.16.1 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
+	github.com/sirupsen/logrus v1.9.4 // indirect
 	go.yaml.in/yaml/v2 v2.4.2 // indirect
+	go.yaml.in/yaml/v3 v3.0.4 // indirect
 	golang.org/x/sys v0.42.0 // indirect
 	google.golang.org/protobuf v1.36.11 // indirect
 	modernc.org/libc v1.70.0 // indirect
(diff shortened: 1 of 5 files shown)


juev (approved):
Thanks for closing the residual round-trip gap in the GHSA-7hp6 hardening — gating literal-block style behind `literalBlockSafe` with a double-quoted fallback is the right fix, and well-formed PEM stays readable. The `FuzzGenerate` harness that reloads through Nebula's real `config.C` parser is a valuable addition; the new go.mod entries are all indirect and already in the module graph via Nebula. LGTM.

PR #161, opened 2026-05-25
fix(web): unify host-ownership anchor on host.CAID across edit/update/mobile-bundle

### LINKED ISSUE #156: web: unify host-ownership anchor on host.CAID across edit/update/mobile-bundle
## Background

The GHSA-c6v2-3ffm-vcmc fix introduced `loadAccessibleHost` (internal/web/handlers.go), which gates `handleHostDetail` / `handleHostBlock` / `handleHostDelete` on the host's own `ca_id`, mirroring the JSON API's `canAccessHost` → `actorOwnsCA(h.CAID)`.

Three handlers were out of scope for that advisory and still gate ownership on **`network.CAID`** instead:

- `handleHostEdit`
- `handleHostUpdate`
- `handleGenerateMobileBundle`

## Why it matters

There is **no live cross-tenant gap today**: both create paths set `host.CAID = networkCAID(network)`, so the two anchors agree. But the divergence is fragile — any future path that lets `host.CAID` differ from its network's CA (e.g. a default-CA fallback) would make these three handlers authorize differently from detail/block/delete and from the API.

## Proposed change

Route the three handlers through `loadAccessibleHost` (or an equivalent `host.CAID` check) so `host.CAID` is the single authoritative ownership anchor across the web layer, matching the API. Add regression tests covering non-owner 403 on edit/update/mobile-bundle with store-state assertions.

Surfaced during review of the GHSA-c6v2-3ffm-vcmc fix.

juev:
## Summary

Unifies the web layer's host-ownership anchor on `host.CAID`, matching the JSON API.

`handleHostEdit`, `handleHostUpdate`, and `handleGenerateMobileBundle` previously gated ownership on **`network.CAID`** (the CA of the host's *network*). The other host handlers (`handleHostDetail` / `handleHostBlock` / `handleHostDelete`, fixed under GHSA-c6v2-3ffm-vcmc) and the entire JSON API (`canAccessHost` → `actorOwnsCA(h.CAID)`) gate on **`host.CAID`** (the CA of the *host itself*).

These three handlers are now routed through the existing `loadAccessibleHost` helper, making `host.CAID` the single authoritative ownership anchor across the web layer.

There is no live cross-tenant gap today — both create paths set `host.CAID = networkCAID(network)`, so the two anchors agree. This removes the fragile divergence: any future path where `host.CAID` differs from its network's CA would otherwise make these handlers authorize differently from detail/block/delete and from the API.

## Changes

- `internal/web/handlers.go`:
  - `handleHostEdit` — replace the manual `CurrentOperator`/`GetHost`/`network.CAID` ownership block with `loadAccessibleHost`; keep the `GetNetwork` load for CIDR display.
  - `handleHostUpdate` — keep `ParseForm` first and `op` (still needed for the `op.Username` audit entry); replace the ownership block with `loadAccessibleHost`.
  - `handleGenerateMobileBundle` — `loadAccessibleHost` first, then the mobile-kind gate.
- `internal/web/tenant_scope_test.go` — three regression tests (non-owner 403 on edit/update/mobile-bundle, plus owner-positive and store-state assertions).

Net: ~92 lines of duplicated ownership logic removed, 3 tests added.

## Side effects (both hardening)

- **`handleGenerateMobileBundle` now checks ownership before the mobile-kind gate** — a non-owner gets `403` instead of `400`, so they can no longer probe another tenant's host kind.
- **Closes a latent nil-operator skip** — the old `if op != nil && op.Role != "admin"` guard let an unauthenticated request bypass the ownership check entirely; `loadAccessibleHost` redirects nil operators to login.

## Testing

- `go build ./...` — OK
- `go test -race ./internal/web/` — green (new tests + existing admin happy-path tests unaffected; admin branch of `loadAccessibleHost` keeps them passing)
- `go vet ./internal/web/` — clean
- `golangci-lint run ./internal/web/` — identical 23 G120 baseline as `main` (lint-neutral; this change adds no `ParseForm`/`FormValue`)

Closes #156

### DIFF
--- a/internal/web/handlers.go
+++ b/internal/web/handlers.go
@@ -959,41 +959,9 @@ func (w *Web) handleHostDetail(rw http.ResponseWriter, r *http.Request) {
 }
 
 func (w *Web) handleHostEdit(rw http.ResponseWriter, r *http.Request) {
-	op := w.session.CurrentOperator(r)
-	if op == nil {
-		http.Redirect(rw, r, "/ui/login", http.StatusSeeOther)
+	host, ok := w.loadAccessibleHost(rw, r)
+	if !ok {
 		return
-	}
-
-	id := chi.URLParam(r, "id")
-	host, err := w.store.GetHost(r.Context(), id)
-	if errors.Is(err, store.ErrNotFound) {
-		http.NotFound(rw, r)
-		return
-	}
-	if err != nil {
-		w.logger.Error("get host for edit", "error", err)
-		http.Error(rw, "Internal Server Error", http.StatusInternalServerError)
-		return
-	}
-
-	// Ownership check: non-admin operators must own the CA signing the network
-	if op.Role != "admin" {
-		network, err := w.store.GetNetwork(r.Context(), host.NetworkID)
-		if err != nil {
-			w.logger.Error("get network for ownership check", "error", err)
-			http.Error(rw, "Failed to load network", http.StatusInternalServerError)
-			return
-		}
-		if network.CAID == "" {
-			http.Error(rw, "Unauthorized", http.StatusForbidden)
-			return
-		}
-		ca, err := w.store.GetCA(r.Context(), network.CAID)
-		if err != nil || ca.OwnerOperatorID != op.ID {
-			http.Error(rw, "Unauthorized", http.StatusForbidden)
-			return
-		}
 	}
 
 	// Load network for CIDR display and validation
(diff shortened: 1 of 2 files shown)


ak2k (approved):
Reviewed — LGTM, approving. The consolidation routes edit/update/mobile-bundle through the existing loadAccessibleHost helper (unchanged here) rather than adding new authz logic. I confirmed the gate is load-bearing by neutering it locally — all four cross-tenant cases then return 200/400 instead of 403. The host.CAID anchor matches the API's actorOwnsCA/canAccessHost, handleHostUpdate never mutates CAID/NetworkID so it's stable, empty/legacy CAID stays fail-closed for non-admins, and the admin bypass is preserved. tenant_scope_test.go exercises all three unified handlers. Build/-race/lint/gosec clean. Optional follow-up: the update and mobile-bundle tests assert only the foreign-denied path; an own-host allowed assertion would round out the symmetry — non-blocking.

Issue #93, opened 2026-05-13
self-registered operator can create networks/hosts without owning a CA


juev:
## Problem

A self-registered operator (created via `/ui/register`, role=`user`) can create networks and hosts even though they own no CA. As an admin I can see in the DB that the resulting `networks.ca_id` / `hosts.ca_id` columns are either empty or point at the seed admin's CA — neither is what the per-operator CA model (ADR 0002 / issue #46) promises.

Two questions in one:

1. **Should the user be allowed to create a network without owning a CA at all?** Today it silently succeeds. If the design requires every network to be tied to a CA the operator owns, this should fail with a clear "create a CA first" error and a link to `/ui/cas/new`.
2. **Should self-registration create a default CA for the new operator automatically?** Right now `handleRegister` (`internal/web/handlers.go:176`) only creates the operator row. A "create CA on first login" or "create CA on first network create" flow would be friendlier than the current dead-end.

## Repro

1. `/ui/register` as a fresh user.
2. Log in.
3. Navigate to `/ui/networks` → click *New Network* → fill name + CIDR → create.
4. Network appears in the list. `/ui/cas` for the same user shows nothing.
5. Create a host in that network. The host gets a cert — signed by… which CA? Verify in `nebula.db` what `hosts.ca_id` resolved to.

## Where

- `internal/web/handlers.go` — `handleNetworkCreate` (~line 778), `handleHostCreate` (~line 614). Neither calls `loadAccessibleCA`-style logic that `internal/web/cas.go` already uses for the CA detail / retire / delete paths.
- `internal/web/handlers.go` — `handleRegister` (~line 176) is the natural place to seed a per-operator CA if we go with auto-create.

## Notes

Surfaced during local end-to-end testing on 2026-05-13. The session ended with a `lighthouse-1` host whose `ca_id` I never traced — would be useful to confirm whether the existing one-CA-per-deployment fallback is intentional in this code path or a bug.

PR #162, opened 2026-05-25
fix(web): scope accessibleHosts to owned CAs in SQL to fix list undercount past 1000

### LINKED ISSUE #157: web: scope accessibleHosts in SQL via HostFilter.CAIDs to fix list undercount past 1000
## Background

`accessibleHosts` (internal/web/handlers.go) calls `ListHosts(ctx, HostFilter{Limit: 1000})` and then filters to the operator's owned CAs **in Go**. This is the same scope-after-limit pattern that #154 just fixed on the JSON API side.

## Why it matters

Once a deployment has more than 1000 hosts, the global `ORDER BY name LIMIT 1000` can fill the window with hosts under CAs the operator does **not** own, dropping the operator's own hosts before the Go ownership filter runs. A non-admin then silently sees fewer hosts than they own (undercount). Not a cross-tenant leak — a correctness/visibility bug.

## Proposed change

Port the API fix: pass `HostFilter.CAIDs` (added in #154) so the owned-CA scope is applied in SQL **before** `ORDER BY name / LIMIT`, then drop the Go post-filter. The code comment at internal/web/handlers.go (the `accessibleHosts` doc comment) already flags this follow-up.

Related: #154.

juev:
## Problem

`accessibleHosts` (`internal/web/handlers.go`) called `ListHosts(ctx, HostFilter{Limit: 1000})` and then filtered to the operator's owned CAs **in Go**. This is the same scope-after-limit pattern #154 fixed on the JSON API side.

Once a deployment has more than 1000 hosts, the global `ORDER BY name LIMIT 1000` can fill the window with hosts under CAs the operator does **not** own, dropping the operator's own hosts before the Go ownership filter runs. A non-admin then silently sees fewer hosts than they own (undercount). Not a cross-tenant leak — a correctness/visibility bug. The doc comment on `accessibleHosts` already flagged this follow-up.

Closes #157.

## Fix

Port the #154 fix into `accessibleHosts`:

- Non-admins: compute owned CA ids and pass them as `HostFilter.CAIDs`, so the owned-CA scope is applied in SQL (`AND ca_id IN (SELECT value FROM json_each(?))`) **before** `ORDER BY name / LIMIT`. Short-circuit to `[]` when the actor owns no CAs — an empty `CAIDs` means "unscoped" in `ListHosts`, so the guard prevents a CA-less non-admin from seeing every host.
- The Go post-filter (`owned` map + loop) is removed.
- Admins are unchanged (no `CAIDs` → unscoped, still capped at 1000 — the global pagination cap is a separate concern).

Legacy blank-`ca_id` hosts match no owned CA id, so the SQL scope keeps them hidden from non-admins (issue #93 behavior preserved). The store layer (`HostFilter.CAIDs`, SQL scoping in `ListHosts`) already landed in #154 — this PR is handler-only.

All three callers (`handleDashboard`, `handlePartialStats`, `handleHosts`) benefit through the single shared helper.

## Tests

- `TestWebHosts_NonAdminNotUndercountedPastLimit`: with 1001 foreign hosts sorting ahead of the owner's one host, a non-admin still sees their host via `GET /ui/hosts` — the web-side reproduction (mirrors the API-side test).
- Existing tenant-scope suite stays green (`TestWebHosts_ListScopedToOwner`, `TestWebHosts_AdminSeesAllTenants`, `TestWebHost_EmptyCAIDHiddenFromNonAdmin`).

`go test ./... -race` is green; `golangci-lint run ./internal/web/...` reports only the pre-existing G120 baseline (diff is lint-neutral).

Related: #154.

### DIFF
--- a/internal/web/handlers.go
+++ b/internal/web/handlers.go
@@ -563,37 +563,30 @@ func computeStats(hosts []*models.Host, networkCount int) dashboardStats {
 }
 
 // accessibleHosts returns the hosts an operator may see: admins see every
-// host; non-admins see only hosts under CAs they own. Mirrors
-// accessibleNetworks (a blank ca_id is a legacy pre-multi-CA row, hidden from
-// non-admins per issue #93). The owned-CA filter runs in Go after the store
-// query; once HostFilter.CAIDs lands (the API handleListHosts fix, work-yzu3)
-// this should scope in SQL so the row cap can't undercount.
+// host; non-admins see only hosts under CAs they own. The owned-CA scope is
+// pushed into the query via HostFilter.CAIDs so it applies before ORDER BY /
+// LIMIT — a global LIMIT applied before the ownership filter could drop an
+// operator's own hosts out of the window, silently undercounting once the
+// deployment exceeds the cap (#157, mirrors the API handleListHosts fix in
+// #154). A blank ca_id is a legacy pre-multi-CA row: it matches no owned CA
+// id, so the SQL scope hides it from non-admins per issue #93.
 func (w *Web) accessibleHosts(ctx context.Context, op *models.Operator) ([]*models.Host, error) {
-	hosts, err := w.store.ListHosts(ctx, store.HostFilter{Limit: 1000})
-	if err != nil {
-		return nil, err
-	}
-	if op.Role == "admin" {
-		return hosts, nil
-	}
-	cas, err := w.store.ListCAsByOwner(ctx, op.ID)
-	if err != nil {
-		return nil, err
-	}
-	owned := make(map[string]struct{}, len(cas))
-	for _, c := range cas {
-		owned[c.ID] = struct{}{}
-	}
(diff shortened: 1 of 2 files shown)


ak2k (approved):
Reviewed — LGTM, approving. I confirmed the fix by reverting accessibleHosts to the old Go-side filter (the new test fails with the undercount) and restoring it (passes): the owned-CA scope is now applied in SQL before ORDER BY/LIMIT, so a non-admin's host beyond row 1000 is returned. Isolation is sound — ca_id IN (SELECT value FROM json_each(?)) binds the CA IDs as one JSON-array parameter, no injection. Admin path is unscoped, zero-owned-CAs returns empty early (no empty-IN leak), and op is nil-guarded. TestWebHosts_NonAdminNotUndercountedPastLimit seeds 1001 sort-leading foreign hosts and asserts both no-undercount and no-leak, and fails against the old code. Build/-race/lint/gosec clean.

PR #163, opened 2026-05-25
docs(adr): 0009 scale, concurrency, and fuzz testing

ak2k:
ADR 0009 — the testing-direction write-up you asked me to commit. It records the
three-tier plan and pins the PR-gate vs scheduled CI boundary so it doesn't erode:

- Fast lane (gates PRs): unit + `-race` + Tier 1 fuzz **seed corpus** + the
  deterministic Tier 2 simulation invariants.
- Scheduled/manual lane (off the PR path): generative `-fuzztime` fuzzing, long
  simulation runs, and Tier 3 Nebula interop.

Tier 1's first piece — `FuzzGenerate` and the GHSA-7hp6 round-trip fix it found —
merged in #155. The Tier 2 framework (the simulation harness + the clock seam)
follows in its own PR, structured to this boundary. Happy to adjust the specific
calls here before it lands.

### DIFF
--- a/docs/adr/0009-scale-and-fuzz-testing.md
+++ b/docs/adr/0009-scale-and-fuzz-testing.md
@@ -1,0 +1,256 @@
+# ADR 0009 — Scale, concurrency, and fuzz testing
+
+- **Status**: Proposed
+- **Date**: 2026-05-25
+- **Context**: The test suite has strong per-request unit coverage but no layer
+  that exercises the control plane at fleet scale, under concurrency, or
+  against adversarial input. The direction here (upstream all three tiers) is
+  agreed; this ADR pins the specific CI-gating calls.
+
+## Context
+
+The repository's tests are predominantly table-driven unit tests plus an
+in-process integration harness (`tests/integration/`) that stands up the real
+API server behind `httptest.NewServer` over an in-memory SQLite store and signs
+real proof-of-possession (PoP) polls (`signedGetUpdates`, ADR 0004). CI runs
+`go test -race` and `golangci-lint`.
+
+What this does **not** cover:
+
+- **Fleet scale.** No test drives the control plane with thousands of hosts.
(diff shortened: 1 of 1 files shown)


juev:
Read through the ADR alongside the code it references, and the technical detail holds up well — `FuzzGenerate` and its `am_lighthouse`/`punchy.punch`/`listen.port` round-trip, the `GetBlocklist` + `has_updates` behavior in `updates.go`, the config-version pair, migration 014's `host_addresses`, and the `cawatch`/`alerts` scanners all match what's in the tree. The three-tier framing is clear, and I like that the PR-gate vs scheduled boundary is spelled out as normative rather than left implicit.

One thing worth fixing before it lands, plus two small notes:

**The clock seam is marked "(implemented)" but isn't in the tree yet.** The Tier 2 section presents `Server.WithClock`, `pki.ShouldRenewAt`, and `SignRequest.Now` as done ("Behavior is unchanged when the clock is unset…"), but I couldn't find any of them in the code — the `pki` package and `SignRequest` are there, but not the `Now` field or `ShouldRenewAt`. That also reads against the PR description and the Phasing section, which both say the clock seam follows in its own PR, and against the overall `Status: Proposed`. Since the doc calls itself normative, the mismatch is the kind of thing a future reader trips over. I'd switch `(implemented)` to `(proposed)` and put that paragraph in the future tense.

**The uniqueness constraint is 018, not 014.** `IP_UNIQUENESS` credits migration 014 with normalizing addresses into `host_addresses` and says "a network-scoped uniqueness constraint backs this" — but the constraint itself came in `018_host_address_network_uniqueness`. Worth pinning 018 there.

**README index.** `docs/adr/README.md` doesn't list 0009 — though it already stops at 0005, so this isn't a regression. Might be a good moment to add 0009 and backfill 0006–0008, but that's easily a follow-up.

Otherwise this reads well — it's documentation only, nothing touching a code path.

PR #169, opened 2026-05-26
docs(adr): 0009 review follow-ups (clock-seam status, migration 018, ADR index)

ak2k:
Post-merge follow-up to #163, addressing the three points from your review there.

1. **Clock seam marked `(proposed)`, not `(implemented)`.** `Server.WithClock`, `pki.ShouldRenewAt`, and `SignRequest.Now` aren't in the tree yet — the seam lands in its own PR (per the Phasing section). I switched the label and put the paragraph in the future tense so it's consistent with `Status: Proposed` and the phasing note.
2. **`IP_UNIQUENESS` now credits migration 018.** Migration 014 normalized addresses into `host_addresses`; the network-scoped uniqueness constraint itself came in `018_host_address_network_uniqueness`. Fixed the attribution.
3. **ADR index backfilled.** `docs/adr/README.md` stopped at 0005 — I added 0006–0009.

Docs only; no code paths touched.

### DIFF
--- a/docs/adr/0009-scale-and-fuzz-testing.md
+++ b/docs/adr/0009-scale-and-fuzz-testing.md
@@ -104,14 +104,15 @@ tests; the simulation earns its place on the interleavings a single-feature
 test won't reach — revocation racing renewal under load, that class of thing —
 and each fix gains a Tier 2 invariant as it lands.
 
-**Clock seam (implemented).** Cert renewal, the rotation overlap window, and
+**Clock seam (proposed).** Cert renewal, the rotation overlap window, and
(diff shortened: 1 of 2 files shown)


juev (approved):
Thanks, Adam — clean docs follow-up. The migration 018 attribution for `IP_UNIQUENESS` and the README index backfill (0006–0009) are both correct.

One ordering tweak from me: since #170 just landed the clock seam, I pushed a small commit restoring the **"Clock seam (implemented)"** framing on this branch (present tense, pointing at #170) so the ADR matches the tree once this merges. Everything else is yours, unchanged.

PR #170, opened 2026-05-27
test(simtest): Tier-2 fleet-simulation harness + clock seam (ADR 0009)

ak2k:
Implements ADR 0009 Tier 2: an in-process fleet-simulation harness plus the injectable clock seam it needs. ADR 0009 already merged (#163); this lands the code conforming to the CI split it pins.

## Commits (reviewable in order)

1. **`feat(api,pki): injectable clock seam`** — *please review this one in focus; it touches the auth-adjacent poll path.* Adds `api.Server.WithClock(func() time.Time)` (default `time.Now`, behavior unchanged when unset), `pki.ShouldRenewAt(nb, na, now)` with `ShouldRenew` delegating, and an optional `pki.SignRequest.Now`. The signed-poll handler routes its renewal decision, rotation-overlap window, and timestamp-skew check through the seam, and threads `now` into `signHostCert`. Recorded/forensic timestamps (audit, revocation `At`/`BlockedAt`) intentionally stay on the wall clock.
2. **`test(simtest): fleet-simulation harness + invariants`** — N virtual agents drive the real `httptest` server under `-race`; named invariants are checked against the store + an event journal after each step: `CONFIG_CONVERGENCE`, `CERT_RENEWAL`, `REVOCATION_LIVENESS`, `TENANT_ISOLATION`, `TOKEN_SINGLE_USE`, `IP_UNIQUENESS`, plus scale characterization.
3. **`test(api,pki): deterministic seam unit tests`** — direct unit coverage for the seam (fixed-instant `ShouldRenewAt` table, exact-instant `SignRequest.Now`, handler-level renewal-window polls).
4. **`test(simtest): review-pass polish`** — small test-honesty fixes.
5. **`test(simtest): activate ConcurrentCreate IP_UNIQUENESS invariant`** — migration 018 (#149) is on main, so this previously-skipped invariant is now an active regression guard over that fix (verified stable across repeated `-race` runs).

## CI split (per the ADR)

The deterministic invariants run in the existing `go test -race ./...` gate (whole sim suite is well under a second). Nothing generative or long-running is on the PR path. The scheduled lane (generative fuzz with a time budget, long sim runs, Tier-3 interop) is a separate follow-up — this PR honors the boundary without wiring that workflow yet.

## Notes

- The `cawatch`/`alerts` scanners still read `time.Now()` directly — seaming them is the follow-up the ADR calls out.
- `last_seen_at` is routed through the seam by design (deterministic under simulation); it's never set in production, where it remains `time.Now()`.
- The harness lives under `internal/simtest/` and is imported by nothing outside its own tests.

Verified: `go test -race ./...` green; `make lint` (pinned) reports 0 issues.

### DIFF
--- a/internal/api/hosts.go
+++ b/internal/api/hosts.go
@@ -445,7 +445,7 @@ func (s *Server) handleRotateCert(w http.ResponseWriter, r *http.Request) {
 		writeError(w, http.StatusInternalServerError, "no current cert to re-sign")
 		return
 	}
-	signed, err := s.signHostCert(r.Context(), host, certInfo)
+	signed, err := s.signHostCert(r.Context(), host, certInfo, s.now())
 	if err != nil {
 		s.logger.Error("sign cert for rotate-cert", "error", err)
 		writeError(w, http.StatusInternalServerError, "failed to sign new cert")
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -34,6 +34,21 @@ type Server struct {
 	limiter            *ratelimit.Limiter
 	passwordPolicy     auth.Policy
 	enrollmentTokenTTL time.Duration
+	clock              func() time.Time // nil => time.Now; injectable for tests
+}
+
+// WithClock overrides the server's wall clock. The default (nil) uses
+// time.Now. Tests inject a controllable clock to exercise time-dependent paths
+// (cert renewal, the rotation overlap window, poll timestamp skew) without
+// real waits. Behavior is unchanged when unset.
+func (s *Server) WithClock(fn func() time.Time) { s.clock = fn }
+
+// now returns the current time from the injected clock, or time.Now.
+func (s *Server) now() time.Time {
+	if s.clock != nil {
+		return s.clock()
+	}
+	return time.Now()
 }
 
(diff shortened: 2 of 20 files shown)


juev (approved):
Thanks, Adam — really clean landing.

A few things I appreciated on review:

- The seam stays behavior-preserving: clock unset ⇒ `time.Now()` at every call site, so production is untouched.
- Keeping forensic timestamps (audit `At`, revocation `BlockedAt`) on the wall clock while routing functional decisions (skew, rotation-overlap, renewal, `last_seen`, token TTL) through `now` is exactly the right split.
- The simtest invariants are honestly documented — the `ipupdate` NOTE that a PASS is a post-fix regression guard, not proof the TOCTOU is closed, is the kind of test honesty I wish I saw more often.

I ran the concurrency invariants (`TokenSingleUse`, `IPUniqueness`) 5× under `-race` plus the full simtest/api/pki suite locally — all green.

One tiny non-blocking note: the description says `last_seen_at` is "never set in production", but it is set — to `time.Now()` (since `now == s.now()` when the clock is unset). The code is correct; only the wording is slightly off.

PR #172, opened 2026-05-27
test: Tier-1 fuzz backlog — PoP canonical, IP validation, migrations, enroll, agent-poll (ADR 0009)

ak2k:
## What

Five native-Go fuzz targets (ADR 0009 Tier 1), seeded from the existing table tests. **Test-only — no production code changes.**

| Target | Package | Property |
|--------|---------|----------|
| `FuzzPoPCanonical` | `internal/pop` | canonical-string injectivity (separator-free inputs) + agent-`Sign`/server-`Verify` agreement, incl. wrong-key and tamper rejection (ADR 0004 §7.1) |
| `FuzzValidateIP` | `internal/api` | `validateHostIPs` accept ⟹ sound: each IP parses, is in a configured CIDR, is not an IPv4 network/broadcast address, unique, and free of any existing-host collision |
| `FuzzMigrations` | `internal/store` | migration-chain totality (no panic / no half-apply) + idempotency + 014 `nebula_ip`→`host_addresses` backfill conservation, on schema-plausible legacy data |
| `FuzzEnrollPayload` | `tests/integration` | `POST /api/v1/enroll` answers every arbitrary body with a client error — never 5xx, never 2xx |
| `FuzzAgentUpdatesHeaders` | `tests/integration` | signed-poll PoP verify core (signature decode, canonical assembly, ed25519 verify, RFC3339 timestamp, nonce-replay cache) — never 5xx, never 2xx |

Each target's seed corpus runs as an ordinary test in the existing `-race` lane; the generative pass runs on the scheduled slow lane added in #171.

## Notes

- The two integration targets stand the real `httptest` server up once. To allow that, I widened the shared test helpers (`setupE2E`, `apiCall`, `signingKeypair`, `enrollHost`) from `*testing.T` to `testing.TB` — backward-compatible, since `*testing.T` satisfies it — and added a small `provisionNetwork` plus an `enrollHostForFuzz` adapter over the existing `enrollHost`.
- Building these surfaced two minor robustness gaps that I'll send as focused follow-up PRs, each paired with the fuzz target that proves it: (1) creating a host with a duplicate name returns 500 instead of 409, and (2) `validateHostIPs` doesn't apply the IPv4 network/broadcast reservation to the IPv4-mapped IPv6 (`::ffff:…`) form of an address.
- Validated locally: `go test -race ./...` green, `golangci-lint` clean, each target explored with a short `-fuzztime` (no crashers).

### DIFF
--- a/internal/api/validate_ip_fuzz_test.go
+++ b/internal/api/validate_ip_fuzz_test.go
@@ -1,0 +1,116 @@
+package api
+
+import (
+	"context"
+	"net/netip"
+	"testing"
+
+	"github.com/forgekeep/nebula-mesh/internal/models"
+	"github.com/forgekeep/nebula-mesh/internal/store"
+)
+
+// fuzzIPStore is a minimal store.Store for FuzzValidateIP. validateHostIPs
+// touches only GetNetwork and ListHosts, so every other method is left to the
+// nil embedded interface — calling one would panic and surface an unexpected
+// new dependency instead of silently passing.
+type fuzzIPStore struct {
+	store.Store
+	net   *models.Network
+	hosts []*models.Host
+}
+
+func (f *fuzzIPStore) GetNetwork(context.Context, string) (*models.Network, error) {
+	if f.net == nil {
+		return nil, store.ErrNotFound
+	}
+	return f.net, nil
+}
+
+func (f *fuzzIPStore) ListHosts(context.Context, store.HostFilter) ([]*models.Host, error) {
+	return f.hosts, nil
+}
+
+// FuzzValidateIP drives validateHostIPs with random network CIDR sets and
+// requested addresses (ADR 0009 Tier 1). The safety property is
+// one-directional: whenever validation *accepts*, that acceptance must be
+// sound — every address parses, lands inside one configured CIDR, is not an
+// IPv4 network/broadcast address, is unique within the request, and collides
+// with no existing host. A false accept is an overlay-IP collision or an
+// out-of-range address reaching the data layer. The dual-family case (a
+// network carrying both a v4 and a v6 CIDR) falls out of fuzzing both CIDRs
+// independently. The target must also never panic.
+//
+//	go test ./internal/api/ -run '^$' -fuzz='^FuzzValidateIP$'
+func FuzzValidateIP(f *testing.F) {
+	f.Add("10.0.0.0/24", "fd00::/64", "10.0.0.5", "fd00::5", "10.0.0.9")
+	f.Add("10.0.0.0/30", "", "10.0.0.0", "10.0.0.3", "")          // network + broadcast (reserved)
+	f.Add("192.168.0.0/16", "", "192.168.1.1", "192.168.1.1", "") // duplicate within request
+	f.Add("", "", "", "", "")
+
+	f.Fuzz(func(t *testing.T, cidr1, cidr2, ip1, ip2, existingIP string) {
+		netCIDRs := fuzzNonEmpty(cidr1, cidr2)
+		reqIPs := fuzzNonEmpty(ip1, ip2)
+
+		st := &fuzzIPStore{net: &models.Network{ID: "net1", CIDRs: netCIDRs}}
+		if existingIP != "" {
+			st.hosts = []*models.Host{{ID: "other", Name: "other", NebulaIPs: []string{existingIP}}}
+		}
+
+		// A rejection — or a malformed-CIDR/network error — is always allowed;
+		// only an *acceptance* carries a soundness obligation.
+		if err := validateHostIPs(context.Background(), st, "net1", reqIPs, ""); err != nil {
+			return
+		}
+
+		// Accepted: re-derive the invariant independently. validateHostIPs only
+		// returns nil when every CIDR parsed, so this prefix list matches the
+		// one it used. Note: this re-check deliberately reuses forbiddenIPv4Reserved
+		// and the same Is4() family gate, so it asserts the accept path is
+		// self-consistent — a bug rooted in those shared helpers (e.g. how a
+		// 4-in-6 address is classified) would pass both sides by construction.
+		prefixes := make([]netip.Prefix, 0, len(netCIDRs))
+		for _, c := range netCIDRs {
+			if p, perr := netip.ParsePrefix(c); perr == nil {
+				prefixes = append(prefixes, p)
+			}
+		}
+		seen := make(map[string]bool, len(reqIPs))
+		for _, ip := range reqIPs {
+			addr, perr := netip.ParseAddr(ip)
+			if perr != nil {
+				t.Fatalf("accepted an unparseable IP %q (cidrs=%v)", ip, netCIDRs)
+			}
+			var containing netip.Prefix
+			inRange := false
+			for _, p := range prefixes {
+				if p.Contains(addr) {
+					containing, inRange = p, true
+					break
+				}
+			}
+			if !inRange {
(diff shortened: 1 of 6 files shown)


juev (changes requested):
Test-only PR — reviewed the full diff, cross-checked the asserted invariants against the production code (`validateHostIPs`, `Migrate`, `handleEnroll`, `decodeSigningPublicKeyPEM`), and ran all five seed corpora locally (green). Solid work; the property statements are precise and the scoping of the two robustness follow-ups is honest. One concrete fix before merge, plus a couple of non-blocking observations.

## 🔴 Change requested

### `enroll_fuzz_test.go` — the "well-formed" seed never reaches its documented depth

```go
if _, signingPub, err := ed25519.GenerateKey(rand.Reader); err == nil {
    signingPEM := pem.EncodeToMemory(&pem.Block{Type: "NEBULA ED25519 PUBLIC KEY", Bytes: signingPub})
```

`ed25519.GenerateKey` returns `(PublicKey, PrivateKey, error)`, so `signingPub` is bound to the **private key (64 bytes)**, not the public key. The PEM then wraps 64 bytes under a "PUBLIC KEY" label.

In `handleEnroll` the third guard is `decodeSigningPublicKeyPEM`, which rejects anything where `len(block.Bytes) != ed25519.PublicKeySize` (32). The 64-byte blob fails there, so enroll returns **400 at the signing-PEM guard and never reaches the enrollment-token lookup**.

That contradicts the seed's own comment ("clears the early guards and reaches enrollment-token lookup, which 401s"). The seed currently just duplicates the other malformed inputs. The test still passes (400 satisfies never-5xx/never-2xx), so the defect is silent — exactly the failure mode fuzz seeds are meant to avoid.

Fix — take the first return value:

```go
if signingPub, _, err := ed25519.GenerateKey(rand.Reader); err == nil {
```

With a real 32-byte key the PEM clears the guard, and `token="bogus-but-well-formed"` drives `GetEnrollmentToken` → `ErrNotFound` → 401, as documented.

## 🟡 Non-blocking observations

- **`migrations_fuzz_test.go` — `applyMigrationsBefore` re-implements `Migrate`.** It execs each migration file whole (`ExecContext(string(b))`), whereas `Migrate` uses `splitSQLStatements` plus per-statement `duplicateColumnErr` tolerance and the repair path. It works today because the modernc driver runs all statements in one Exec and a fresh `:memory:` DB never hits the duplicate-column/repair case (confirmed: pre-014 schema builds correctly, seeds green). Low risk, but it's a drift point — if the loader's statement-splitting semantics change, or the driver is swapped for one that runs only the first statement per Exec, this helper silently builds a partial schema.
- **`validate_ip_fuzz_test.go` re-check is a self-consistency oracle, not an independent one.** It reuses `forbiddenIPv4Reserved` and the same `Is4()` gate, so a bug rooted in those helpers (e.g. 4-in-6 `::ffff:…` classification) passes both sides by construction. Already documented in the comment/commit, with the independent 4-in-6 oracle tracked as a follow-up — flagging only so the limitation is explicit in review.
- **`FuzzMigrations` conservation is count-based**, so it's robust to address normalization during backfill but won't detect it. Consistent with the stated property; just noting the boundary.

## What's good

- `*testing.T` → `testing.TB` widening is correctly backward-compatible (only `Helper`/`Fatal`/`Cleanup`, no `t.Run`/`Parallel`).
- `fuzzIPStore` with a nil-embedded `store.Store` panics on any unexpected method — surfaces a new dependency instead of passing silently. Nice.
- Honest scoping: the dup-name 500→409 and 4-in-6 reserved gaps are split into separate follow-up PRs paired with the targets that prove them, rather than folded into a test-only PR.
- Seed-corpus (fast `-race` lane) vs. generative pass (slow lane from #171) split matches ADR 0009.

After the enroll-seed fix this is good to merge.

ak2k:
Fixed in b2654ab.

`signingPub` now takes the first return of `ed25519.GenerateKey` (the public key, 32 bytes), so the well-formed seed clears `decodeSigningPublicKeyPEM` and reaches the token lookup → `GetEnrollmentToken` → `ErrNotFound` → 401, as the comment claims. I checked the byte lengths directly: the old binding produced a 64-byte blob (the private key), the fix produces 32.

You're right that this was silent — 400 and 401 both satisfy the never-5xx/never-2xx assertion, so a green run proved nothing about the seed's depth. To stop that recurring I extracted the payload into a shared `wellFormedEnrollSeed` helper (so the corpus and the assertion can't drift) and added `TestEnrollWellFormedSeedReaches401`, which asserts the seed returns 401. I reintroduced the private-key binding to confirm the guard bites: the new test fails with `got 400, want 401`, so the degradation is now a hard failure instead of a silent one.

I left the three non-blocking observations as-is — the `applyMigrationsBefore` drift and the 4-in-6 self-consistency limitation are honest follow-ups, and the count-based conservation matches the stated property.

ak2k:
Thanks for the careful review — and for catching the enroll seed; it's a clean example of a seed silently failing to reach the depth it claims.

**Enroll seed (fixed in `b2654ab`):** switched to the first return value (`signingPub, _, err := ed25519.GenerateKey(...)`) so the PEM carries the 32-byte public key, clears `decodeSigningPublicKeyPEM`, and the bogus token then drives `GetEnrollmentToken` → `ErrNotFound` → 401 as documented. To stop that silent-degradation class from recurring I extracted the seed into a shared `wellFormedEnrollSeed` builder and added `TestEnrollWellFormedSeedReaches401`, which asserts the seed reaches 401 — the coarse never-5xx/never-2xx fuzz assertion can't tell a 400 guard from the intended 401, so this pins the depth explicitly.

**On the non-blocking observations:**

- `applyMigrationsBefore`: agreed it re-derives the loader and relies on the driver running all statements per `Exec` on a fresh `:memory:` schema that never hits the duplicate-column/repair path. I can add a comment marking that assumption, or route it through the loader's statement-splitter so the harness tracks `Migrate` exactly — happy to do either in a quick follow-up; your call.
- `validateHostIPs` re-check: yes, self-consistency by construction; the independent 4-in-6 oracle (and the reserved-check fix it guards) is the separate follow-up.
- `FuzzMigrations` conservation is count-based by design — won't detect address normalization during backfill, matching the stated property.

CI is green; ready for another look when you have a moment.

juev (approved):
Fix in b2654ab is exactly what was needed: `signingPub, _, err := ed25519.GenerateKey(...)` takes the public half (32 bytes), the seed clears `decodeSigningPublicKeyPEM` and reaches token lookup → 401 as documented. Verified locally — all five seed corpora pass under `-race`.

The guard-test addition (`TestEnrollWellFormedSeedReaches401` + shared `wellFormedEnrollSeed` helper) is the right shape: it pins the depth that the coarse never-5xx/never-2xx assertion cannot see, and prevents corpus/assertion drift. Good response to a class of bug, not just to this instance.

The three non-blocking observations are honest follow-ups — agreed to keep them separate. LGTM.

juev:
Thanks — and especially for how you responded to the seed-depth issue. Fixing the immediate bug (public vs private half of the keypair) was the obvious move; extracting `wellFormedEnrollSeed` into a shared helper and adding `TestEnrollWellFormedSeedReaches401` so the seed and the assertion can't drift apart was the less-obvious one, and that's what fixes the *class* of bug — silent seed degradation under a coarse never-5xx/never-2xx assertion — not just the instance. The honest scoping of the dup-name 500→409 and 4-in-6 reserved gaps as separate follow-ups paired with the targets that prove them is exactly the right call for a test-only PR.

PR #173, opened 2026-05-27
ci: cap the scheduled fuzz job with timeout-minutes (follow-up to #171)

ak2k:
Follow-up to your optional, non-blocking suggestion on #171: cap the scheduled fuzz job with `timeout-minutes: 30`. `workflow_dispatch` accepts an arbitrary `fuzztime`, so the only ceiling today is GitHub's 6h job default — a typo'd or oversized dispatch could pin a runner. 30m sits comfortably above the nightly 10m/target + setup while bounding a runaway. `actionlint` clean.

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -39,6 +39,10 @@ jobs:
   fuzz:
     name: generative fuzz (${{ matrix.target }})
     runs-on: ubuntu-latest
+    # Safety ceiling: workflow_dispatch accepts an arbitrary fuzztime, so cap the
+    # job well below GitHub's 6h default — a typo'd or oversized fuzztime can't
+    # pin a runner. Comfortably above the nightly 10m/target plus setup.
+    timeout-minutes: 30
     strategy:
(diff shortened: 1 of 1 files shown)


ak2k:
Closing without merging — on a closer look I don't think the cap is worth adding.

Walking the failure modes a `timeout-minutes` guard would catch:

- A normal run is already bounded by `-fuzztime` itself; the test stops when the budget elapses, so job duration ≈ fuzztime + setup overhead.
- A hung or wedged process is bounded by GitHub's 6h job default.
- A stuck slot is freed by the existing `concurrency: cancel-in-progress` group — the next scheduled or dispatched run supersedes a hung one.
- The one case a tighter cap catches early is a typo'd `workflow_dispatch` fuzztime — but a tight cap (30m here) also clips the intentional deep fuzz runs the dispatch input exists to enable.

With free Actions removing the cost driver too, the marginal value didn't justify shipping it. Appreciate the original nudge — your "fine to land as-is" framing flagged it as optional, and I'm landing on the "don't add it" side of that.

PR #175, opened 2026-05-28
fix(store): actionable migration-018 startup error on overlay-IP conflicts

ak2k:
## Summary

Follow-up to #149. Migration 018 (`UNIQUE(network_id, address)` over
`host_addresses`) is intentionally fail-loud on a pre-existing duplicate overlay
IP, but an operator hitting it today sees only the raw runner breadcrumb
(`apply migration 018_... stmt "CREATE UNIQUE INDEX ...": <sqlite error>`),
which names neither the offending network/IP/hosts nor the remediation. The
operator whose server won't boot is reading the startup log, so I made the
failure self-explanatory there.

I added a pre-apply guard in `Migrate` (mirroring the existing
`CountEmptyCAIDRows` startup check) that runs before 018 and handles its two
reachable failure modes by class:

- **Cross-host duplicate** (two or more distinct hosts share one overlay IP) — a
  security defect (a CA has issued certificates that let one host receive the
  other's traffic) with no data-derivable answer to which host should win. Fails
  loud with a message naming each conflict, and leaves the database untouched.
- **Orphan `host_addresses` rows** (referenced host no longer exists) and
  **same-host position duplicates** (one host bound to the same address twice) —
  redundant/dangling data with no policy decision and no certificate
  implication. Repaired automatically and logged via `slog.Warn`.

## Why this shape

018 has exactly two reachable failure modes on a valid schema: the backfill
`UPDATE` hits a `NOT NULL` violation on an orphan row, and `CREATE UNIQUE INDEX`
fails on a duplicate. A guard that checks both has complete coverage, so I kept
it a pure pre-check rather than wrapping the statement-exec loop — which means a
conflicted database that fails loud is left byte-identical ("no schema changes
were made"), and the only case requiring a human is the one where resolution
genuinely depends on operator knowledge of which host owns the IP.

The remediation wording is deliberately precise about certificates: removing a
`host_addresses` row (or reassigning the address) does not revoke the displaced
host's certificate — only deleting the host through the management API/UI does
(`DeleteHostAndBlockCert`). Since the management server is down during this
failure, the message directs the operator to clear the conflict at the database
level to boot, then revoke the certificate once the server is running, and notes
that resolving it before upgrading avoids the failure entirely.

## Tests

`internal/store/migration_018_test.go`: cross-host fail-loud (including that the
fail path mutates nothing even when cleanable rows are present), same-host
auto-collapse (preserving a host's legitimately distinct second address), orphan
auto-removal, grouping/dedup, message truncation caps, cleanup audit logging,
and two end-to-end `Migrate` tests over a real store. The existing direct-SQL
`TestMigration018_FailsLoudOnDuplicateOverlayIP` is unchanged.

### DIFF
--- a/internal/store/migration_018_test.go
+++ b/internal/store/migration_018_test.go
@@ -1,8 +1,12 @@
 package store
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
+	"fmt"
+	"log/slog"
+	"path/filepath"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


juev (approved):
Reviewed end-to-end (sqlite.go guard, test file, Migrate integration) and ran the full test set under `-race` locally. All twelve tests green.

## What I checked

- **Classification + ordering**: cross-host fails loud *before* any cleanup runs — verified directly by `TestCheckOverlayIPConflicts_CrossHostFailDoesNotCleanup`, which seeds a cross-host conflict alongside orphan and same-host-duplicate rows and asserts the row count is unchanged on the fail path. The "DB byte-identical on fail" claim holds.
- **Idempotency**: the guard sits inside the `if applied { continue }` loop, so a successful first run skips it on every subsequent start; a fail-loud run doesn't record 018 in `schema_migrations`, so the operator's post-resolution restart re-runs the guard cleanly. Neither case double-mutates.
- **Cross-host detection**: `JOIN hosts h ON h.id = ha.host_id` correctly excludes orphan rows from the cross-host query (an orphan can't be a "host claiming an IP"). The cleanup that runs *after* the cross-host check then handles the orphans.
- **Operator message**: precise about `DeleteHostAndBlockCert` being the only revocation path, and the two-phase "clear at DB level → boot → revoke via management UI" wording is coherent given the server is down during the failure.
- **Truncation caps**: both group-count cap and per-group host-count cap are tested at `maxConflictsListed+3` and `+5` respectively.

## Non-blocking observations

1. **Orphan cleanup can silently mask a historical cross-host overlap.** If host B was removed from `hosts` (FK off) while its `host_addresses` row remained, and host A is still bound to the same IP, the cross-host JOIN excludes B's row → no conflict reported → orphan cleanup deletes B's row → 018 succeeds. The `slog.Warn` audit line records the orphan removal but doesn't flag the overlap with a live host. Acceptable for the hand-edited/FK-off recovery scenario the doc scopes to, just worth being explicit about — "this guard does not reconstruct conflicts whose other side is already gone from `hosts`."
2. `crossHostOverlayConflicts` uses an O(N²) correlated subquery. Fine for typical installs; only mentioning because someone reading the SQL later might wonder. Not worth pre-emptive rework.
3. `errors.New(formatCrossHostConflicts(...))` doesn't wrap an underlying SQL error — intentional, the message itself names "migration 018" so context isn't lost. No change needed.

Good defensive engineering, the test coverage of "fail path mutates nothing" is the right invariant to pin. LGTM.

juev:
Thanks for the framing here. The cross-host vs same-host/orphan policy split — security defect with no data-derivable winner ⇒ fail loud and don't touch the DB; vs redundant/dangling rows ⇒ auto-repair and log — is the right shape, and pinning the "fail path mutates nothing" invariant with `CrossHostFailDoesNotCleanup` (cleanable rows seeded *alongside* the conflict) makes it self-enforcing against future edits rather than relying on argument. The cert-revocation precision in the operator message — that removing a `host_addresses` row does NOT revoke, only `DeleteHostAndBlockCert` does — turns a panic-inducing startup error into something an operator can act on without making the security posture worse.

Issue #126, opened 2026-05-20
configgen: marshal agent config via yaml.v3 typed struct (follow-up to GHSA-7hp6)


juev:
## Background

GHSA-7hp6-g3pq-3pc3 (PR #126 — private fork) closed the immediate YAML-injection vector by tightening input validation on `advanced.tun_device`. The advisory's longer-term recommendation was to eliminate the entire injection class by replacing the `text/template`-based agent config generator with typed-struct marshalling through `gopkg.in/yaml.v3`.

This issue tracks that follow-up.

## Why this is still worth doing after the input-side fix

`internal/configgen/generator.go` currently builds `config.yml` by string-interpolating user-supplied values into a `text/template` literal (`configTemplate`, lines 61–172). That is safe today only because every interpolated field is validated upstream:

- `ListenHost` → `ValidateIPAddr` (IP only, no metacharacters possible)
- `TunDevice` → `[A-Za-z0-9_-]{1,15}` whitelist (after GHSA-7hp6 fix)
- `UnsafeRoutes.Route/Via` → `ValidateCIDR` / `ValidateIPAddr`

The risk shape this leaves: any future advanced field (or any change to existing validators that loosens them) re-opens the injection class. Defense in depth at the output side removes that footgun.

## Proposed change

Replace `configTemplate` + `template.Execute` with a typed `nebulaConfig` Go struct marshalled via `gopkg.in/yaml.v3`:

```go
type nebulaConfig struct {
    PKI           pkiSection     `yaml:"pki"`
    StaticHostMap map[string][]string `yaml:"static_host_map,omitempty"`
    Lighthouse    lighthouseSection   `yaml:"lighthouse"`
    Listen        listenSection       `yaml:"listen"`
    Punchy        punchySection       `yaml:"punchy"`
    Tun           *tunSection         `yaml:"tun,omitempty"`
    Relay         *relaySection       `yaml:"relay,omitempty"`
    Logging       loggingSection      `yaml:"logging"`
    Firewall      firewallSection     `yaml:"firewall"`
}
```

yaml.v3 quotes / escapes any value safely, so a `TunDevice` of `nebula0\rinjected: true` becomes the literal string `"nebula0\rinjected: true"` in the output rather than a YAML structural break.

## Scope

- New file `internal/configgen/marshal.go` with the typed struct + a `Generate` reimplementation
- Delete `configTemplate`, `indentLines`, `template.FuncMap`
[...]

PR #177, opened 2026-06-01
fix(configgen): round-trip operator strings in generated YAML (#176)

### LINKED ISSUE #176: configgen: FuzzGenerate crasher — operator strings with a tab break generated YAML
## Summary

The scheduled generative-fuzz lane (`Scheduled` workflow, job `generative fuzz (FuzzGenerate)`) failed on run [26709587640](https://github.com/forgekeep/nebula-mesh/actions/runs/26709587640). This is a genuine bug the fuzzer caught, not CI flakiness.

```
fuzz_test.go:81: generator emitted invalid YAML:
  yaml: line 7: found a tab character where an indentation space is expected
```

## Root cause

`internal/configgen` marshals the Nebula config from a typed struct through `gopkg.in/yaml.v3`. The inline-PEM fields are already guarded against non-round-tripping output (`literalString` + `literalBlockSafe`, GHSA-7hp6 / #126), but the **plain operator-controlled string fields** were not — `static_host_map` (keys and values), `lighthouse.hosts`, `listen.host`, `tun.dev`, firewall rules, and `relays` were emitted as bare `string`/`[]string`.

For a value such as a lighthouse `PublicAddr` of `"\t\n"`, yaml.v3's encoder auto-selects literal-block style (`|`) for the multi-line content and emits the tab verbatim under the block indent. yaml.v3's own parser — the same path a Nebula agent runs on SIGHUP reload — then rejects it (`found a tab character where an indentation space is expected`).

The header comment in `marshal.go` claimed every value is quoted/escaped so structural-break characters "can never escape into the YAML structure" — that claim was false for the auto-block-style case.

### Reproduce

```
go test ./internal/configgen/ -run '^$' -fuzz='^FuzzGenerate$'
```

Crashing input: lighthouse `PublicAddr = "\t\n"` (fields otherwise trivial).

## Fix

Introduce a `safeString` type (the plain-string counterpart to `literalString`) whose `MarshalYAML` forces double-quoted style for multi-line values — the only style yaml.v3 auto-selects that breaks round-trip — while leaving single-line values to the encoder so ordinary values stay unquoted. Apply it to every operator-controlled scalar field, including `static_host_map` keys. Add the crasher as an `f.Add` regression seed so it runs in the normal (fast-lane) `go test`.

## Scope

- `internal/configgen/marshal.go`
- `internal/configgen/fuzz_test.go`

juev:
## What

Fixes the `FuzzGenerate` crasher the scheduled fuzz lane caught in run [26709587640](https://github.com/forgekeep/nebula-mesh/actions/runs/26709587640).

A lighthouse `PublicAddr` of `"\t\n"` reached `static_host_map` as a bare string, where yaml.v3 auto-selected literal-block style (`|`) and emitted the tab verbatim under the block indent. yaml.v3's own parser — the SIGHUP-reload path a Nebula agent runs — then rejected it:

```
yaml: line 7: found a tab character where an indentation space is expected
```

## Why

The inline-PEM fields were already guarded against non-round-tripping output (`literalString` + `literalBlockSafe`, GHSA-7hp6 / #126), but the plain operator-controlled string fields were not — `static_host_map` (keys and values), `lighthouse.hosts`, `listen.host`, `tun.dev`, firewall rules, and `relays` were emitted as bare `string`/`[]string`. The `marshal.go` header comment claimed these "can never escape into the YAML structure"; that was false for the auto-block-style case.

## How

- Add a `safeString` type — the plain-string counterpart to `literalString` — whose `MarshalYAML` forces double-quoted style for multi-line values (the only style yaml.v3 auto-selects that breaks round-trip) and leaves single-line values to the encoder, so ordinary values stay unquoted and readable.
- Apply it to every operator-controlled scalar field, including `static_host_map` keys.
- Update the misleading header comment.
- Record the crasher as an `f.Add` regression seed so it runs in the normal `go test`, not only the scheduled fuzz lane.

## Verification

- `go build ./...`
- `go test -race ./internal/configgen/...` (incl. new seed)
- `make lint` — 0 issues
- 30s live fuzz, 370k execs, no new crashers

Closes #176

### DIFF
--- a/internal/configgen/fuzz_test.go
+++ b/internal/configgen/fuzz_test.go
@@ -48,35 +48,20 @@ func FuzzGenerate(f *testing.F) {
 	// round-tripping style for such values.
 	f.Add("0", "0", "0", "\t\n", "0", "0", "0", "0", "0", -125, 74,
 		false, true, false, false, "0")
+	// Regression for the invalid-UTF-8 case: safeString now tags scalars
+	// "!!str", and yaml.v3 refuses to marshal invalid UTF-8 under that tag.
+	// A safeString field carrying a lone 0xFF byte must fall back to a
+	// "!!binary" node (as a native Go string would) rather than fail Generate.
+	f.Add("\xff", "\xff", "\xff", "\xff", "\xff", "\xff", "\xff", "\xff", "\xff",
+		0, 0, false, false, true, false, "")
 
 	f.Fuzz(func(t *testing.T,
 		hostName, ip, lhIP, lhAddr, listenHost, tunDev, fwPort, fwProto, fwGroup string,
 		listenPort, mtu int, isLH, isRelay, punch, inlinePEM bool, pemBlock string,
 	) {
-		punchy := punch // address-of a copy; PunchyOverride wants *bool
-		in := GeneratorInput{
-			HostName:         hostName,
-			NebulaIPs:        []string{ip},
-			IsLighthouse:     isLH,
-			IsRelay:          isRelay,
-			ListenPort:       listenPort,
-			ListenHost:       listenHost,
-			MTU:              mtu,
-			TunDevice:        tunDev,
-			PunchyOverride:   &punchy,
-			Lighthouses:      []LighthouseInfo{{NebulaIPs: []string{lhIP}, PublicAddr: lhAddr}},
-			FirewallInbound:  []FirewallRule{{Port: fwPort, Proto: fwProto, Group: fwGroup}},
-			FirewallOutbound: []FirewallRule{{Port: fwPort, Proto: fwProto, Group: fwGroup}},
-		}
-		// Generate only emits the inline-PEM section when CACertPEM is set
-		// (see buildConfig); otherwise it falls back to the path form.
-		if inlinePEM && pemBlock != "" {
-			in.CACertPEM, in.CertPEM, in.KeyPEM = pemBlock, pemBlock, pemBlock
-		} else {
-			in.CACertPath = "/etc/nebula/ca.crt"
-			in.CertPath = "/etc/nebula/host.crt"
-			in.KeyPath = "/etc/nebula/host.key"
-		}
+		in := fuzzGeneratorInput(hostName, ip, lhIP, lhAddr, listenHost, tunDev,
+			fwPort, fwProto, fwGroup, listenPort, mtu, isLH, isRelay, punch,
+			inlinePEM, pemBlock)
 
 		out, err := Generate(in)
 		if err != nil {
@@ -99,8 +84,8 @@ func FuzzGenerate(f *testing.F) {
 		if got := c.GetBool("lighthouse.am_lighthouse", !isLH); got != isLH {
 			t.Fatalf("am_lighthouse round-trip: got %v want %v\n--- config ---\n%s", got, isLH, out)
 		}
-		if got := c.GetBool("punchy.punch", !punchy); got != punchy {
-			t.Fatalf("punchy.punch round-trip: got %v want %v\n--- config ---\n%s", got, punchy, out)
+		if got := c.GetBool("punchy.punch", !punch); got != punch {
+			t.Fatalf("punchy.punch round-trip: got %v want %v\n--- config ---\n%s", got, punch, out)
 		}
 		wantPort := listenPort
 		if wantPort == 0 && isLH {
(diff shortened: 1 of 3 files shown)


ak2k (approved):
Nice catch by the scheduled fuzz lane, and the fix reads well — forcing double-quoted style for multi-line values is exactly right for the `\t\n` literal-block tab, and seeding it as an `f.Add` regression is great. A few suggestions to round it out, all in the same round-trip spirit `safeString` is going for.

**A small comment tweak.** The doc mentions single-line values being "auto-quoted for ambiguous scalars like `0`" — I think that part may have gotten slightly ahead of the code: since the node is built tagless (`Style:0`), it loses the implicit `!!str` tag a native Go `string` carries, so ambiguous scalars actually emit bare. Quick check against the pinned yaml.v3 v3.0.1:

| input | this PR | native `string` (pre-PR) | re-parses as |
|---|---|---|---|
| `"0"` | `0` | `"0"` | int |
| `"true"` | `true` | `"true"` | bool |
| `"null"`/`"~"` | `null`/`~` | quoted | null |
| `""` | bare | `""` | null |

The practical effect is just that ambiguous operator values render bare now (e.g. `port: 22` where it used to be `port: "22"`) — might be worth a small wording update either way.

**A couple of values we could round-trip while we're here.** `static_host_map` is `map[safeString][]safeString`, and the key marshaler fires too, so a few edge values don't quite survive a round-trip through a typed loader:

| key | this PR | result |
|---|---|---|
| `null`/`~` | `null:` | key decodes to nil → entry dropped |
| `<<` | `<<:` | `map merge requires map or sequence of maps` → load fails |
| `""` value | — | decodes to `[]`, not `[""]` |

All gated in practice — keys are validated overlay IPs and the firewall path stays fail-closed — so nothing urgent. But since `safeString` is aiming for round-trip "for any string," it'd be nice to close them.

**An easy way to get there** — tag the node `!!str` and keep the multi-line branch:

```go
func (s safeString) MarshalYAML() (any, error) {
	n := &yaml.Node{Kind: yaml.ScalarNode, Tag: "!!str", Value: string(s)}
	if strings.Contains(string(s), "\n") {
		n.Style = yaml.DoubleQuotedStyle
	}
	return n, nil
}
```

That restores the pre-PR rendering (plain when safe, quoted only for ambiguous scalars), mirrors what `literalString` already does, and picks up the `null`/`~`/`""` cases. `!!str` on its own doesn't cover the multi-line crash (the encoder still picks a literal block for `\t\n`), so the `\n` branch still earns its place. One it doesn't catch is a `<<` *key*, which stays a fail-safe parse error on an input that can't really occur — easy to just mention in the comment. If you'd rather not reason about that at all, unconditional `Style: yaml.DoubleQuotedStyle` closes everything in one go, at the cost of quoting every scalar.

**Maybe one more test.** The `\t\n` seed is doing real work — it fails pre-fix — though it only runs through `lighthouse.PublicAddr`, and the invariant type-checks only the non-`safeString` scalars, so a future regression on a `safeString` field could still pass green. Could be worth a small focused test that shares the generator-input builder with the fuzz body, pushes a value like `"\t\n"`/`"0"`/`""` through `Generate` → `config.LoadString`, and checks it reads back unchanged (and that the multi-line case stays double-quoted rather than a block). `unsafe_routes`/`relays`/PKI paths aren't covered for round-trip yet either.

Happy to push the one-liner + test up if that's easier than folding it in.

juev:
Thanks for the thorough review, Adam — went ahead and folded in all of it. Pushed as `96bb3ff`.

**`!!str` tag.** Adopted, and confirmed against pinned yaml.v3 v3.0.1 that it restores the pre-PR rendering: ambiguous scalars (`0`/`true`/`null`/`~`/`""`) auto-quote again instead of bare-emitting and re-typing on load (so `port: "22"` is a string once more). Comment rewritten to match — the old "auto-quoted for ambiguous scalars" line is now accurate rather than aspirational.

**The `<<` key.** Closed it rather than just documenting — `!!str` alone leaves bare `<<:` as the merge indicator (verified it still fails the loader), so the marshaler now also forces double-quoted style for `<<`. While verifying I also caught that a **multi-line key** hits a second block path: yaml.v3 emits it as an explicit-key `? |` block, so the `\n` branch earns its place for keys too, not only values.

**One thing the `!!str` tag surfaced.** The scheduled fuzz lane immediately found that tagging the node `!!str` makes yaml.v3 reject invalid UTF-8 (`cannot marshal invalid UTF-8 data as !!str`) — a regression the tagless version didn't have. Fix is to tag only valid UTF-8; invalid sequences stay tagless and fall back to a `!!binary` node, exactly as a native Go `string` would, which round-trips. Locked in with an `f.Add("\xff", …)` seed. Final marshaler:

```go
func (s safeString) MarshalYAML() (any, error) {
	n := &yaml.Node{Kind: yaml.ScalarNode, Value: string(s)}
	if utf8.ValidString(string(s)) {
		n.Tag = "!!str"
		if strings.Contains(string(s), "\n") || string(s) == "<<" {
			n.Style = yaml.DoubleQuotedStyle
		}
	}
	return n, nil
}
```

**The focused test.** Added `TestGenerate_SafeStringRoundTrip`. It reuses the extracted `fuzzGeneratorInput` builder (shared with the fuzz body, as you suggested, so it can't drift) and pushes `0`/`true`/`null`/`~`/`""`/`<<`/`\t\n`/`a:b`/`# x` through every safeString field — `listen.host`, `static_host_map` key **and** value, `firewall.port`, `relay.relays`, `unsafe_routes.route/via`, and the `pki` path — asserting each reads back unchanged through `config.LoadString`, with a `mustString` helper that fails if a value silently re-typed (the `"22"`→`int` case). A separate test pins that the multi-line value renders double-quoted on one line, not as a block.

`go test -race ./...`, `make lint`, and a 20s fuzz run are all green. Mind taking another look?

ak2k:
Took another look at `96bb3ff` — this is clean, all four points folded in well, and the bonus catch is the best part.

The `!!str` + invalid-UTF-8 → `!!binary` fallback is the right call: I confirmed it renders `host: !!binary /w==` for a lone `0xFF` byte and round-trips, where an explicit `!!str` tag would have errored. `<<` key forced to double-quoted, multi-line value/key stays double-quoted (the test pins the rendering, not just the round-trip), and `TestGenerate_SafeStringRoundTrip` reusing `fuzzGeneratorInput` keeps the focused checks from drifting off what the fuzzer drives.

One thing I chased down on the `!!str` quoting, since forcing string style on `firewall.port` could in theory change semantics: it does not. Nebula reads every rule scalar through `convertRule` → `fmt.Sprintf("%v", v)` (`firewall.go`), so bare `port: 22` and quoted `port: "22"` both coerce to the same `"22"` before `parsePort`. The change is invisible to the firewall parser.

Re-checked against the merged commit: `go test -race ./internal/configgen/...`, the focused round-trip + multiline-rendering tests, a 15s live fuzz (no crashers), and `golangci-lint` are all green. Happy with it.

PR #231, opened 2026-06-13
chore: bump Go toolchain to 1.26.4

ak2k:
govulncheck flags two stdlib vulnerabilities reachable from this code,
both fixed in go1.26.4:

- GO-2026-5039 (net/textproto): reachable via io.ReadAll ->
  ReadMIMEHeader in internal/cli/user.go
- GO-2026-5037 (crypto/x509 hostname parsing): reachable via
  ListenAndServe certificate verification in internal/cli/serve.go and
  internal/cli/network.go

After the bump, govulncheck ./... reports no reachable vulnerabilities.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.7
 
-ARG GO_VERSION=1.26.3
+ARG GO_VERSION=1.26.4
 
 FROM golang:${GO_VERSION}-alpine AS build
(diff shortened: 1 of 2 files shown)


juev:
Thanks for this — clean govulncheck-driven bump, and the body documents exactly which two stdlib advisories (GO-2026-5039, GO-2026-5037) are reachable and why. Merging.

PR #232, opened 2026-06-13
fix(ratelimit): key on rightmost X-Forwarded-For entry, not leftmost

ak2k:
With trust_proxy_header enabled, ClientIP returned the first parseable
X-Forwarded-For entry. The comment claimed the leftmost public address
wins, but there was no public/trust check — and the leftmost entry is
the one the client itself sends. A reverse proxy appends the connecting
address on the right (nginx's $proxy_add_x_forwarded_for, the shipped
snippets in deploy/reverse-proxy/), so an attacker could rotate a fake
leftmost value per request to evade the per-IP auth, enroll, and
agent_poll buckets, and poison the masked IP recorded in audit entries.

Trust the rightmost entry instead: it is the only one written by the
proxy we trust. There is deliberately no leftward fallback when the
rightmost entry fails to parse — that would reopen the same bypass —
the limiter falls back to RemoteAddr instead.

### DIFF
--- a/internal/ratelimit/limiter.go
+++ b/internal/ratelimit/limiter.go
@@ -112,20 +112,21 @@ func (l *Limiter) Allow(ip, group string) (allowed bool, retryAfter time.Duratio
 }
 
 // ClientIP returns the effective IP for the request. When the limiter
-// is configured to trust X-Forwarded-For, the leftmost public address
-// wins; otherwise RemoteAddr is used.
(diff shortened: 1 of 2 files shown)


juev:
Great catch. The leftmost X-Forwarded-For entry being client-controlled made the whole per-IP limiter bypassable, and trusting the rightmost (proxy-appended) entry with no leftward fallback is the right model. Merging.

Issue #180, opened 2026-06-01
hardening: constant-time login to prevent username enumeration


juev:
Found during the security audit (#178).

`internal/web/session.go:73-86`: `Login` returns immediately for a non-existent username, but runs `bcrypt.CompareHashAndPassword` (~100ms) when the user exists. The timing difference allows statistical username enumeration despite rate limiting.

**Proposal:** always perform a dummy `bcrypt.CompareHashAndPassword` against a fixed placeholder hash when the user is not found, so both paths take comparable time.

PR #233, opened 2026-06-13
fix(web): equalize login timing for OIDC accounts

ak2k:
OIDC-provisioned operators store the "oidc" sentinel instead of a
bcrypt hash, so the password compare in Login fails instantly with
ErrHashTooShort — no KDF work. The #180 dummy-hash equalizer only
covered the unknown-username and inactive-account paths, leaving a
timing oracle that distinguishes active OIDC accounts (microseconds)
from wrong passwords and unknown usernames (~full bcrypt cost).

Run the dummy compare whenever the stored hash is unusable (any compare
error other than ErrMismatchedHashAndPassword), and extend the
constant-time login test with an OIDC-account case. The new assertion
fails against the previous code (16.8µs vs a ~23ms floor on this
machine) and passes with the fix.

### DIFF
--- a/internal/web/login_timing_test.go
+++ b/internal/web/login_timing_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http"
 	"net/http/httptest"
+	"sort"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


juev:
Nice — the #180 equalizer missed the OIDC sentinel-hash path, leaving a real timing oracle for active OIDC accounts. Keying the dummy compare off any unusable-hash error (not just unknown user / inactive) closes it, and the test pins the floor. Merging.

PR #234, opened 2026-06-13
fix(store,api,web): empty operator role defaults to user, not admin

ak2k:
CreateOperator silently promoted an empty Role to admin. No current
caller hits that branch (register, the operator form, and OIDC
provisioning all set an explicit role), but any future caller that
forgets to set Role would self-provision an administrator — the same
silent-default shape as the empty-username bug fixed in #223.

- CreateOperator: empty role now defaults to user (least privilege)
  and unknown roles are rejected. SeedInitialAdminOperator keeps the
  admin default — it only runs against an empty operators table, and
  the first operator must be able to administer the server — and gains
  the same unknown-role rejection.
- POST /api/v1/operators previously persisted req.Role verbatim; it now
  validates against the known roles and returns 400 for anything else,
  so a typo can't land on a store-level default.
- The operator-create form validates the client-controlled role value
  instead of surfacing the store error as a 500.
- OIDCConfig.Validate rejects a default_role outside {admin,user} at
  startup, so a custom/typo default_role fails loudly instead of 500ing
  every OIDC provisioning login against the new store allowlist.
- New models.OperatorRoleAdmin/OperatorRoleUser constants and
  models.ValidOperatorRole.

The API-key lifecycle test now creates its operator with an explicit
admin role; it exercises admin-gated endpoints and previously relied on
the implicit admin default.

### DIFF
--- a/internal/api/operators.go
+++ b/internal/api/operators.go
@@ -38,6 +38,15 @@ func (s *Server) handleCreateOperator(w http.ResponseWriter, r *http.Request) {
 		writeError(w, http.StatusBadRequest, "username and password are required")
 		return
 	}
+	// An omitted role means least privilege; anything else must be a known
+	// role so a typo can't silently land on a store-level default.
+	if req.Role == "" {
+		req.Role = models.OperatorRoleUser
+	}
+	if !models.ValidOperatorRole(req.Role) {
+		writeError(w, http.StatusBadRequest, "role must be \"admin\" or \"user\"")
+		return
+	}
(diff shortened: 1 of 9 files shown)


juev:
Agreed — silently promoting an empty role to admin is the same footgun as the empty-username bug in #223. Defaulting to user, rejecting unknown roles, and layering validation at store/API/web/OIDC is exactly right, while SeedInitialAdminOperator correctly keeps the admin default. Merging.

Issue #179, opened 2026-06-01
hardening: enforce TLS by default (refuse plaintext HTTP unless explicitly opted out)


juev:
Found during the security audit (#178).

`internal/cli/serve.go:338-342` binds plain HTTP when `tls_cert`/`tls_key` are unset; TLS is only advisory (log warning + docs). All credentials (API keys, passwords, TOTP, mobile bundles) then transit in cleartext.

Per SECURITY.md, "running without TLS in production" is out of scope as a vulnerability — filing this as **hardening**: the insecure posture is the *default*.

**Proposal:** require `tls_cert`+`tls_key` (or an explicit `i_understand_no_tls: true` / `--insecure-http` flag for reverse-proxy setups) before binding plaintext HTTP. Fail-closed by default.

PR #235, opened 2026-06-13
fix(agent): refuse plaintext http:// server_url unless explicitly opted out

ak2k:
The CLI validates --server (#219) and the server refuses cleartext
binds on routable addresses (#179), but the agent accepted any
non-empty server_url. With http://, enrollment POSTs the single-use
token in cleartext and every poll fetches config_yml + certificate_pem
over an unauthenticated channel — an on-path attacker can steal the
token, harvest the certificate, or inject a malicious config.yml that
the agent writes to disk and SIGHUPs Nebula to load.

ValidateAgentServerURL requires https, allowing http only for loopback
hosts (local testing, an on-host TLS-terminating proxy) — deliberately
narrower than the server's private-range allowance, since a private
address still crosses a real network. The explicit opt-out mirrors the
server: allow_insecure_http in agent.yml or --insecure-http on the
daemon and enroll commands.

Both LoadAgentConfig and SaveAgentConfig validate, and the enroll path
checks the flag before agent.Enroll, so a refused URL never carries the
token. httptest servers are loopback, so existing tests pass the guard
unchanged.

An already-enrolled host whose pre-guard config carries an http:// URL
would otherwise park silently in standby after a binary upgrade; the
daemon now logs the rejection at Warn with the remediation so it does
not read as a routine not-enrolled-yet state.

### DIFF
--- a/cmd/nebula-agent/enroll_test.go
+++ b/cmd/nebula-agent/enroll_test.go
@@ -269,6 +269,32 @@ func TestEnrollSubcommand_RequiresServerAndToken(t *testing.T) {
 	}
 }
 
+// TestEnrollSubcommand_RefusesPlaintextHTTP pins the https-required guard on
+// the enroll path: a cleartext non-loopback --server is refused before the
+// single-use token leaves the process (no config or key files are written).
+// httptest URLs are loopback, so every other enroll test passes the guard
+// untouched; --insecure-http opts out.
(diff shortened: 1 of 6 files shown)


juev:
Good defense-in-depth. The CLI and server guards left the agent accepting plaintext server_url, so the token and config still crossed the wire in cleartext. The loopback-only http allowance plus the explicit opt-out mirrors the server, and surfacing the legacy-config rejection at Warn avoids the silent-standby trap. Merging.

Issue #196, opened 2026-06-01
keystore: master key bytes not zeroized after base64 decode in NewMasterFromBase64


juev:
## Summary

`NewMasterFromBase64` never zeroizes the base64-decoded master-key bytes after handing them to `NewMaster`, leaving a plaintext copy of the KEK on the heap.

## Details

`internal/keystore/keystore.go:39-48`:
```go
func NewMasterFromBase64(b64 string) (*Master, error) {
    ...
    raw, err := base64.StdEncoding.DecodeString(b64) // line 43
    if err != nil { ... }
    return NewMaster(raw) // line 47 — raw never wiped
}
```

`raw` goes out of scope un-wiped, so a plaintext copy of the 32-byte master key lingers on the heap until GC reclaims (and possibly reuses) the page. The rest of the codebase zeroizes CA key material aggressively (the recent GHSA hardening), so this is an inconsistent residual copy.

## Suggested fix

```go
raw, err := base64.StdEncoding.DecodeString(b64)
if err != nil { ... }
defer keystore.Zeroize(raw)
return NewMaster(raw)
```

(`NewMaster` / `aes.NewCipher` copies the key into the cipher's key schedule, so wiping `raw` after the call is safe.)

## Severity

Low — defense-in-depth, consistent with existing zeroization policy.

_Found during a full-codebase review (recall-biased)._

Issue #181, opened 2026-06-01
hardening: zeroize decrypted CA key material on LoadByID error path


juev:
Found during the security audit (#178).

`internal/pki/resolver.go:56-63`: when `OpenWithDEK` decrypts `keyBytes` (plaintext ed25519 private key) but `LoadCAFromMaterial` then fails (e.g. corrupted CertPEM), `keyBytes` is returned via the error path without zeroization. The `defer` at line 54 wipes only the DEK. Low real-world risk (requires memory-disclosure-level access), but inconsistent with the project's zeroize discipline (`defer caMgr.Wipe()` elsewhere).\n\n**Proposal:** assign `keyBytes` to a named var and `defer keystore.Zeroize(keyBytes)` right after decryption, so all paths wipe it.

PR #236, opened 2026-06-13
fix(agent): zeroize private keys after enrollment writes them to disk

ak2k:
The server side wipes plaintext CA keys on every return path
(caMgr.Wipe, GHSA-8h84 / #181, #196), but the agent left its X25519
host key and Ed25519 PoP signing key — raw bytes and PEM encodings —
live on the heap after enrollment, exposed to core dumps, swap, and
heap inspection on the agent host for the life of the process.

Collect all four buffers in an enrollmentSecrets container wiped by a
single defer, covering both the success path and the error path where
keys were generated but the server rejected the request. A test
observer hook captures the live buffers so tests can assert they read
as zeros after Enroll returns; removing the wipe call fails both new
tests.

### DIFF
--- a/internal/agent/enroll.go
+++ b/internal/agent/enroll.go
@@ -16,6 +16,8 @@ import (
 
 	"github.com/slackhq/nebula/cert"
 	"golang.org/x/crypto/curve25519"
+
+	"github.com/forgekeep/nebula-mesh/internal/keystore"
 )
 
 // SigningPublicKeyPEMType is the PEM block type used by the agent for its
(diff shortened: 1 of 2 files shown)


juev:
Thanks — the server already wipes CA keys, so leaving the agent X25519/Ed25519 keys live on the heap was an asymmetry worth closing. The single-defer enrollmentSecrets container covering both success and error paths is clean, and the observer-based zero assertion is a solid test. Merging.

PR #237, opened 2026-06-13
fix(web,store): make TOTP codes one-time within their validity window

ak2k:
verifyTOTP accepted the same 6-digit code repeatedly across its ~90s
window (period 30s, skew 1): no consumed-code or last-timestep state
was kept, so an attacker who observed one live code (shoulder-surf,
phish) and held the victim's pending-TOTP cookie could replay it.
Recovery codes were already consumed atomically; TOTP codes were not.
RFC 6238 §5.2 requires the verifier to reject reuse.

Migration 020 adds operators.totp_last_timestep, and
ConsumeOperatorTOTPTimestep advances it with the same atomic
compare-and-set shape as ConsumeOperatorRecoveryCode (UPDATE ... WHERE
totp_last_timestep < ?), so two concurrent logins presenting the same
code cannot both pass. verifyTOTPWithTimestep reports which timestep
matched — totp.ValidateCustom doesn't expose that, so each candidate
step in the ±skew window is checked with a constant-time compare.

The login handler consumes the timestep as part of authentication; a
store failure there now surfaces as a 500 rather than masquerading as a
wrong-code denial. The 2FA-enable handler consumes best-effort so the
setup code cannot double as the next login's second factor. New tests
pin replay rejection (fails against the previous handler), next-window
acceptance, the store-level CAS contract, and the store-error-is-500
path.

### DIFF
--- a/internal/store/migrations/020_operator_totp_timestep.down.sql
+++ b/internal/store/migrations/020_operator_totp_timestep.down.sql
@@ -1,0 +1,7 @@
+-- Down migration: removes per-operator TOTP replay protection.
+--
+-- After downgrade the totp_last_timestep column is gone, so an attacker who
+-- observed a valid TOTP code can replay it within its ~90s acceptance window
+-- (period 30s, skew 1) until the column is restored. Operator rows are
+-- otherwise unaffected.
+ALTER TABLE operators DROP COLUMN totp_last_timestep;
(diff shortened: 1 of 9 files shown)


juev:
Solid fix. TOTP codes being replayable across the ~90s window while recovery codes were already consumed atomically was a real RFC 6238 5.2 gap. The compare-and-set on totp_last_timestep mirrors the recovery-code path so concurrent logins cannot both pass. Merging.

PR #238, opened 2026-06-13
fix(keystore,pki): bind CA-key envelopes to their ca_id with GCM AAD

ak2k:
threat-model.md claimed "GCM tag binds AAD/nonce (envelope attacks
refuted)", but all four Seal/Open calls passed nil AAD: neither the
master->DEK wrap nor the DEK->key-material seal was bound to the owning
CA. An attacker with DB write access (no master key) could copy a
matched (encrypted_key_dek, encrypted_key_material) pair from CA-B's
row into CA-A's; LoadByID decrypted B's key and paired it with A's
cert, and LoadCAFromMaterial never noticed the mismatch — the failure
only surfaced later as certificates no peer validates.

Both envelope layers now bind the owning CA's ID as AAD, so the swap
fails at decrypt. Every mint path (API, web, rotation, auto-provision)
allocates the CA ID before sealing. LoadByID falls back to nil AAD for
envelopes sealed before the binding — which cannot reopen the swap: an
AAD-bound envelope fails both opens under a foreign row, and a legacy
swap is caught by the new public-key consistency check in
LoadCAFromMaterial (private key must match the cert it's paired with).
Legacy envelopes pick up the binding when the CA is next rotated.

threat-model.md now describes the actual mechanism. New tests pin
wrong-AAD rejection at both layers, the cross-row swap, the legacy
nil-AAD load, and the key/cert mismatch refusal.

### DIFF
--- a/docs/security/threat-model.md
+++ b/docs/security/threat-model.md
@@ -107,7 +107,7 @@ Trust zones, from least to most trusted:
 ## 5. Cross-cutting mitigations
 
 - **Transport**: TLS by default; non-loopback plaintext bind refused unless opted in (#179).
-- **Crypto**: AES-256-GCM throughout; GCM tag binds AAD/nonce (envelope attacks refuted); key material zeroized (#181, #196).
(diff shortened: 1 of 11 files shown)


juev:
Excellent work — this is the standout of the batch. The threat-model claimed AAD binding that the nil-AAD Seal/Open calls never actually provided, leaving the cross-row envelope swap open. Binding the owning ca_id at both envelope layers, the safe legacy nil-AAD fallback, and the public-key/cert consistency backstop in LoadCAFromMaterial together make the swap fail closed. Updating threat-model.md to match reality is appreciated. Merging.

PR #239, opened 2026-06-13
fix(api,web): audit network creation and firewall policy updates

ak2k:
Every host, CA, operator, and settings mutation writes an audit entry,
but handleCreateNetwork and handleUpdateFirewall did not (the firewall
handler only logged at info level), contradicting threat-model E1
("mutating actions write an audit-log entry"). An admin could create
networks and rewrite firewall policy with no trail — and firewall
policy is exactly the mutation a forensic reader most wants verbatim.

Add network.create (API + web UI form) and network.firewall.update
(API; the UI has no firewall editor). The firewall entry's details
field carries the full new ruleset JSON. Both actions are registered in
the metrics pre-seed, the known-actions enum, and the audit coverage
sweep, which now drives a firewall update as well.

### DIFF
--- a/internal/api/audit_property_test.go
+++ b/internal/api/audit_property_test.go
@@ -35,7 +35,14 @@ func TestAuditRows_AllProductionPathsConform(t *testing.T) {
 	srv, st := newTestServer(t)
 	ctx := context.Background()
 
+	// network.create — the sweep's own fixture network emits it.
 	netID := createNetwork(t, srv)
+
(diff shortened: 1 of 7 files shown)


juev:
Good — network.create and network.firewall.update were the conspicuous gaps against threat-model E1, and firewall policy is exactly the mutation a forensic reader wants verbatim. Carrying the full ruleset JSON in details and extending the coverage sweep is the right shape. Merging.

PR #240, opened 2026-06-13
fix(store): derive audit-log id from crypto/rand, not UnixNano

ak2k:
AddAuditEntry built its PRIMARY KEY from time.Now().UnixNano(), so two
concurrent writes landing on the same nanosecond collided and the
second INSERT failed — and every production caller fire-and-forgets
this method's error, so the security-relevant row vanished with no
trace. The new concurrent-writers test reproduces the collision
reliably against the old id scheme (UNIQUE constraint failure within a
few hundred parallel writes).

Use 16 random bytes (hex) for the id suffix. The etok_/cert_ UnixNano
patterns elsewhere in this file run inside transactions whose errors
propagate, so they fail loud rather than silently and are left as-is.

### DIFF
--- a/internal/store/sqlite.go
+++ b/internal/store/sqlite.go
@@ -2,7 +2,9 @@ package store
 
 import (
 	"context"
+	"crypto/rand"
 	"database/sql"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


juev:
Nice find. A UnixNano primary key colliding under concurrent writes, combined with every caller fire-and-forgetting the error, meant security-relevant rows could vanish silently. 16 random bytes for the id suffix is the right call, and leaving the in-transaction UnixNano patterns alone is correctly scoped. Merging.

PR #241, opened 2026-06-13
fix(web): require non-admins to pick a network when creating a host

ak2k:
A non-admin POSTing /ui/hosts with an empty network_id skipped the
whole CA-ownership block (issue #93's gate only ran when a network was
chosen) and created a host with CAID="". The creator then couldn't
access their own host — loadAccessibleHost forbids blank CAID for
non-admins — yet the orphan still burned an enrollment token and
appeared in admin dashboards.

Reject an empty network_id up front for non-admin operators, with the
same inline form error the ownership checks use. Admin behavior is
unchanged (admins may still create unattached hosts).


juev:
Agreed — an empty network_id let a non-admin create a CAID-less orphan they could not even access, while still burning a token and polluting dashboards. Rejecting it up front with the same inline form error keeps admin behavior unchanged. Merging.

PR #242, opened 2026-06-13
fix(api): stop auditing pre-auth malformed agent polls

ak2k:
A poll missing any of the four PoP headers wrote a host.auth.failed
audit row before any host lookup or signature verification. The branch
needs no token, fingerprint, or signature, so an unauthenticated
sender could bloat the audit table at line rate, bounded only by the
agent_poll rate-limit bucket (60/s per IP).

Drop the DB write for that branch only; every branch after the
fingerprint lookup still audits, and floods remain visible via the
HTTP request counter (route + status 400). The audit coverage sweep
now drives host.auth.failed through the unknown-fingerprint branch
instead, and a new test pins that the missing-headers poll writes no
audit row.

### DIFF
--- a/internal/api/audit_property_test.go
+++ b/internal/api/audit_property_test.go
@@ -8,7 +8,9 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
+	corepop "github.com/forgekeep/nebula-mesh/internal/pop"
 	"github.com/forgekeep/nebula-mesh/internal/store"
 )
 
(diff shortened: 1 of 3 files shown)


juev:
Good — the missing-PoP-headers branch auditing before any host lookup let an unauthenticated sender flood the audit table at line rate. Dropping the DB write for that branch only, while every post-fingerprint branch still audits and the HTTP counter keeps the flood visible, is the right trade. Merging.

Issue #188, opened 2026-06-01
hardening: validate alerts.webhook_url and oidc.issuer (SSRF / require https)


juev:
From the 2nd-pass security audit (#178).

- **Webhook SSRF** (`internal/alerts/sinks.go:92`): the alert sink POSTs to `alerts.webhook_url` with no scheme/host validation. It can target loopback, RFC1918, or cloud-metadata (169.254.169.254) → blind SSRF / internal pivot.
- **OIDC issuer** (`internal/web/oidc.go:74`): `oidc.NewProvider` fetches the issuer discovery doc/JWKS via the default client; the issuer URL is unvalidated (no https requirement).

Both are **admin-config-only** (no runtime/non-admin path to set these URLs), so exploitability is low — but there is zero validation. Real-world analogs (self-hosted control planes) repeatedly ship SSRF via operator-set webhook/integration URLs.

- Fix: at config load, require `https` for `oidc.issuer`; for `webhook_url` reject non-http(s) schemes and resolve-and-reject loopback/private/link-local hosts (with a documented opt-out for intentional internal sinks).

Severity: Minor (admin-only) — defense-in-depth.

PR #243, opened 2026-06-13
fix(alerts): enforce the webhook SSRF guard at delivery time

ak2k:
validateWebhookURL rejects literal private/loopback/link-local hosts at
config load but deliberately does not resolve DNS, and the sender was a
bare http.Client following redirects freely — so a hostname public at
load could resolve privately at delivery time (DNS rebinding), and a
public endpoint answering 302 could steer the POST at
169.254.169.254 or other internal services. Deployer-trusted input and
a status-code-only response bound the impact; this is defense in depth
(self-documented residual of #188).

Run the private-address check in the dialer's Control hook — after DNS
resolution, on the literal IP being connected to — so rebinding and
every redirect hop (each dials through the same transport) are both
covered, with an explicit 10-hop redirect cap. The check unmaps
IPv4-mapped IPv6 addresses first so ::ffff:0.0.0.0 (which routes to
localhost on connect) and other mapped forms classify by their v4
address; the same Unmap is applied to the config-load hostIsPrivate
guard. The shared predicate is factored into isBlockedWebhookAddr and
unit-tested across both v4 and mapped forms.

The alerts.allow_private_webhook opt-out disables the delivery-time
guard too, keeping one knob for intentional internal sinks. A
caller-supplied HTTPClient keeps its own transport policy, so the
existing httptest-based sink tests are unaffected. The response body is
drained before Close so the keep-alive connection can be reused.

### DIFF
--- a/internal/alerts/sinks.go
+++ b/internal/alerts/sinks.go
@@ -7,9 +7,14 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
+	"net"
 	"net/http"
+	"net/netip"
 	"strconv"
+	"syscall"
 	"time"
 
 	"github.com/forgekeep/nebula-mesh/internal/store"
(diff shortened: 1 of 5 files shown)


juev:
Strong defense-in-depth. The load-time guard deliberately skipped DNS, so rebinding and redirect-to-internal were live. Moving the private-address check into the dialer Control hook (post-resolution, per redirect hop) with the IPv4-mapped unmap and a hop cap closes both, and keeping one opt-out knob is clean. Merging.

PR #244, opened 2026-06-13
fix(store): pin Migrate to one connection so PRAGMA foreign_keys holds

ak2k:
Migrate dispatched each migration statement through the connection
pool, but PRAGMA foreign_keys is connection-scoped and every pool
connection opens with foreign_keys(on) via the DSN. Migration 014
turns the pragma OFF, drops and recreates hosts/networks, then turns
it back ON as separate statements — if the OFF and the DROP TABLE
landed on different pooled connections, the drop ran with FK
enforcement live and cascade-deleted every child row: certificates,
host_addresses (backfilled earlier in the same migration),
enrollment_tokens, cert_alerts.

In practice the sequential single-goroutine Exec almost always reused
the same idle connection, which is why this never fired. The new
dispersal test makes it deterministic with SetMaxIdleConns(0): the
unpinned loader fails it (the dispersal surfaces as an FK-constraint
error on DROP TABLE networks; with seeded child rows the same
mechanism silently wipes them).

Pin one *sql.Conn for the whole Migrate run. The deferred cleanup also
restores foreign_keys=ON before the connection rejoins the pool, so a
migration failing mid-recreate can't leak an enforcement-off
connection; a second test pins that FK enforcement is live after
Migrate. Loader helpers now take a small sqlConn interface satisfied
by both *sql.DB and *sql.Conn.

### DIFF
--- a/internal/store/migration_pinned_conn_test.go
+++ b/internal/store/migration_pinned_conn_test.go
@@ -1,0 +1,143 @@
+package store
+
+import (
+	"context"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/forgekeep/nebula-mesh/internal/store/migrations"
+)
+
+// TestMigrate_PinnedConnection_SurvivesPoolDispersal pins the fix for the
+// connection-scoped PRAGMA hazard in migration 014 (L12, 2026-06-12 audit).
+//
+// 014 runs `PRAGMA foreign_keys = OFF`, drops + recreates hosts/networks,
+// then turns the pragma back ON — as separate statements. The pragma is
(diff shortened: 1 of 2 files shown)


juev:
Sharp diagnosis. PRAGMA foreign_keys being connection-scoped meant migration 014 could drop-and-recreate with FK enforcement live on a different pooled connection and cascade-wipe child rows. Pinning one conn for the whole run plus restoring FK=ON on the deferred cleanup is correct, and the SetMaxIdleConns(0) dispersal test makes the latent bug deterministic. Merging.

PR #245, opened 2026-06-13
fix(api,mobilebundle): apply stored firewall rules to generated configs

ak2k:
Operator firewall rules were accepted, validated, persisted, and echoed
by the management API, and updating them bumped the network config
version to trigger an agent re-render — but the two config-generation
sites (renderHostConfig and the mobile-bundle builder) hardcoded
icmp-inbound / allow-all-outbound and never read the stored policy. So
every agent re-rendered the same hardcoded firewall regardless of what
the operator configured: a policy that looked applied but never reached
a host.

This was an incomplete feature, not a deliberate design — commit
25c6c5b ("add lighthouse config propagation, firewall rules, audit
log") documents the intent ("config version bumped on firewall update",
"agents detect config changes via version comparison") and wired the
whole propagation pipeline; only the final read-the-rules step was
missing. host.Advanced does not override the firewall and Nebula's
firewall is default-deny, so until now every generated mesh accepted
only ICMP inbound with no working way to change it.

Both render sites now resolve the network's stored policy via the new
configgen.FirewallRulesFromJSON. A network with no stored policy keeps
the previous baseline (ICMP inbound / allow-all outbound), so behavior
is unchanged for networks that never configured rules. A stored policy
that is malformed or contains a rule missing a selector — the empty
group is the work-322fy footgun: Nebula rejects a rule with neither
group nor host and fails the whole config on load — falls back to that
same baseline rather than emitting a config the agent cannot load (the
API enroll path logs this at Warn; the loggerless mobile-bundle builder
defaults silently). Such rules are only reachable from pre-#200 stored
data, since the write path now rejects empty groups.

Behavior change for operators: a network with a stored, valid firewall
policy will see it take effect on agents' next poll after upgrade.

### DIFF
--- a/internal/api/enroll.go
+++ b/internal/api/enroll.go
@@ -291,22 +291,20 @@ func (s *Server) renderHostConfig(ctx context.Context, host *models.Host) ([]byt
 		}
 	}
 
+	fwInbound, fwOutbound := s.firewallRulesForGenerator(ctx, host.NetworkID)
+
 	input := configgen.GeneratorInput{
-		HostName:     host.Name,
-		NebulaIPs:    host.NebulaIPs,
-		IsLighthouse: host.IsLighthouse,
-		IsRelay:      host.IsRelay,
-		CACertPath:   "/etc/nebula/ca.crt",
-		CertPath:     "/etc/nebula/host.crt",
-		KeyPath:      "/etc/nebula/host.key",
-		ListenPort:   host.ListenPort,
-		Lighthouses:  lighthouses,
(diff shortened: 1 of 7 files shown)


juev:
Thanks — this closes a feature that looked applied but never reached a host: both render sites hardcoded the firewall and never read the stored policy. Resolving the policy via FirewallRulesFromJSON, keeping the prior baseline for networks with no policy, and falling back to baseline on a malformed rule rather than emitting a config Nebula rejects is exactly right. Merging.

PR #281, opened 2026-07-15
fix: Deletion of pending hosts in web UI

inode64:
## What

Fix pending-host deletion in the web UI.

## Why

HTMX requests could be sent without the CSRF header because the global listener was registered on `document.body` before the body existed. The request was rejected with HTTP 403.

## How

Register the `htmx:configRequest` listener on `document`, which is available while the page head is parsed.
Add regression tests for deleting a pending host with CSRF protection and for the global HTMX listener registration.

## Checklist

- [x] New behaviour has tests

### DIFF
--- a/internal/web/host_detail_test.go
+++ b/internal/web/host_detail_test.go
@@ -58,6 +58,61 @@ func TestHostDetail_HasEditButton(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


juev:
Thank you for contributing to Nebula Mesh, @inode64!

PR #306, opened 2026-07-24
feat(host): introduce combined lighthouse+relay role

inode64:
It may be necessary to have Lighthouse and Relay support on the same Nebula instance.

## Checklist

- [x] New behaviour has tests
- [x]  No breaking API changes, or they are called out above


juev:
thx!

PR #309, opened 2026-07-24
feat(firewall): add per-host inbound firewall rules configuration

inode64:
allow the creation of inbound rules to grant groups access to specific ports

## Checklist

- [x] New behaviour has tests


juev (changes requested):
There is one required change before merge.

`advanced_json` is committed by `UpdateHost`, while resetting `config_version` happens in a separate best-effort write. Both the API and Web handlers only log a failure of that second write and still return success. If an inbound rule is removed and the reset fails, the agent can keep using the old, more permissive configuration indefinitely even though the durable Host state says the rule is gone. This violates `SEC-PERSIST-001`.

Please persist the Host update and `config_version = 0` atomically, for example inside the existing `UpdateHost` transaction, and remove the subsequent best-effort reset calls. Please also add a negative regression test referencing `SEC-PERSIST-001` that forces the transaction to fail and verifies that both the firewall state and config version roll back together.

Everything else looks ready: the accidental `.orig` files are gone, all GitHub checks pass, and I locally verified `go build ./...`, `go vet ./...`, `go test -race ./...`, `make lint`, and `git diff --check`.

juev (approved):
The requested atomic Host update and config version reset are implemented with a SEC-PERSIST-001 rollback regression test. Targeted race tests and vet pass.

juev:
thx!

PR #312, opened 2026-07-24
feat(agent): add nebula_reload_command for service reload customization

inode64:
I am setting up nebula-mesh, and I’ve encountered an issue on my Gentoo installations—using both OpenRC and systemd—where Nebula fails to restart because it cannot locate the PID file. The simplest solution is to run a command that triggers a restart directly from the init system itself.

## Checklist

- [x] New behaviour has tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,32 @@ jobs:
           go-version: '1.26.5'
           cache: true
 
+      # nebula_reload_command's Windows half — the cmd.exe command line and the
+      # taskkill /T timeout path — is code the build alone never executes.
+      # Run the reload hook for real on a Windows host.
+      # -run is scoped because the rest of internal/agent still assumes a Unix
+      # host; widen it as those tests are ported.
+      - name: Test reload command (cmd /C path)
+        shell: pwsh
+        run: |
+          $output = go test -count=1 -v -run 'TestNebulaReloader|TestCappedBuffer' ./internal/agent | Out-String
+          Write-Host $output
+          if ($LASTEXITCODE -ne 0) { throw "reload command tests failed" }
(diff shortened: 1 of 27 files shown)


juev:
Thanks — this addresses a real deployment problem for hosts where Nebula does not expose a usable PID file. The overall approach makes sense: `nebula_reload_command` belongs to the local agent configuration, takes precedence over `nebula_pid_file`, and prevents config acknowledgement when the reload fails.

I think the PR needs changes before merge:

1. The timeout does not terminate the full process tree on Linux. The current CI failure demonstrates this: `sleep 5` continues running after the 200 ms test timeout because `CommandContext` kills the shell but not its child. `CombinedOutput` then waits for the child to close the inherited pipes. Please terminate the entire process group on timeout and add a Linux regression test.

2. `CombinedOutput` stores command output in an unbounded buffer. A noisy or surviving child process can consume memory indefinitely. Please cap the captured output while retaining enough context for the error message.

3. The documented `systemctl restart nebula` use case conflicts with the repository’s systemd unit. `nebula-agent.service` declares `PartOf=nebula.service`, so restarting Nebula also restarts the agent and kills the reload command running inside its control group. The agent may exit before sending the config acknowledgement, receive the same update after restart, and repeat the cycle. This interaction needs either a unit/configuration change or an integration test proving a safe command contract.

The Windows job confirms that the agent builds, but it does not execute the new `cmd /C` path. Please add runtime coverage if Windows support remains part of this PR.

The documentation also needs a small consistency pass: the troubleshooting and force-rotate sections still describe only PID/SIGHUP reloads, and the Ansible role cannot render `nebula_reload_command`.

juev:
thx!

juev (approved):

PR #322, opened 2026-07-27
fix(hosts): re-issue certificates when a host's groups change

inode64:
Groups are carried in the Nebula certificate, and firewall rules select their counterparties by group, so editing a host's groups had no effect on the mesh until the certificate was replaced. This bit hardest in the removing direction: a host dropped from a privileged group kept that access for as long as it held a certificate still claiming membership.

The API PATCH handler and the web UI edit form each kept their own copy of the "which fields are cert-bound" rule, and both omitted groups. They now share models.CertIdentityChanged and store.ApplyHostEdit.

ApplyHostEdit sets pending_rekey on the host struct so it commits inside UpdateHost's transaction alongside the config_version reset (SEC-PERSIST-001). That closes an atomicity gap in the API path, which previously issued a separate SetPendingRekey write and so could persist the new groups while failing to schedule the re-issuance that carries them.

Also adds a Groups column to the web UI and CLI host listings.

## Checklist

- [X] `go test -race ./...` passes
- [X] `golangci-lint run ./...` passes
- [X] New behaviour has tests
- [X] User-facing changes are reflected in README / configs
- [X] No breaking API changes, or they are called out above

### DIFF
--- a/cmd/nebula-agent/main.go
+++ b/cmd/nebula-agent/main.go
@@ -82,6 +82,45 @@ func printUsage(w io.Writer) {
 // not mutate it at runtime.
 var standbyTick = 10 * time.Second
 
+const (
+	rekeyRetryInitialDelay = time.Second
+	rekeyRetryMaxDelay     = time.Minute
+)
+
+// rekeyRetryDelay returns an exponentially increasing, bounded delay for
+// consecutive re-enrollment failures. A new Poller performs its first poll
+// immediately, so this delay is the only throttle between failed rekey
+// attempts.
+func rekeyRetryDelay(consecutiveFailures int) time.Duration {
+	if consecutiveFailures <= 1 {
+		return rekeyRetryInitialDelay
+	}
+
+	delay := rekeyRetryInitialDelay
+	for attempt := 1; attempt < consecutiveFailures && delay < rekeyRetryMaxDelay; attempt++ {
+		if delay > rekeyRetryMaxDelay/2 {
+			return rekeyRetryMaxDelay
+		}
+		delay *= 2
+	}
+	return delay
+}
+
+// waitForRekeyRetry makes the retry delay interruptible by SIGTERM/SIGINT.
+// It deliberately uses a timer instead of Sleep so stopping the agent never
+// waits for a backoff window to elapse.
+func waitForRekeyRetry(ctx context.Context, delay time.Duration) error {
+	timer := time.NewTimer(delay)
+	defer timer.Stop()
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-timer.C:
+		return nil
+	}
+}
+
 // runUnified is the daemon entrypoint. After #88 it never fails fast on a
 // missing config or missing enrollment; instead it parks in standby and
 // re-checks every standbyTick until the operator runs `nebula-agent enroll`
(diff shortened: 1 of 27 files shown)


inode64:
fix(rekey): re-offer a rekey until the agent completes it

A rekey used to be lost the moment its token was minted: the poll handler
cleared pending_rekey there, so an agent that could not finish the
re-enrollment left the host on the superseded certificate while the server
reported nothing pending. Re-requesting the rekey just repeated the cycle,
and for a group change that means a host keeps firewall access its host row
says it no longer has.

pending_rekey now clears only inside the enrollment transaction, alongside
the certificate it asked for. While it is set the poll re-offers the rekey,
minting a fresh token each time — only the hash is stored, so the previous
token cannot be re-sent; CreateTokenForHost already deletes the host's other
unused tokens, so exactly one is ever live. This also makes the
host.rekey.completed audit entry fire, which it never could before: the flag
it keys on was always already cleared by the time enrollment ran.

The agent no longer exits when a rekey fails. It logs and keeps polling, so
the server's next offer retries it. Exiting delegated the retry to the
service manager, which restarts on a fixed delay and eventually trips its
start limit — turning a recoverable failure into an agent stopped for good,
silently applying neither config nor certificate updates.

deploy/systemd/nebula-agent.service gains the two settings that made this
reachable in the first place:

  - ReadWritePaths now covers /etc/nebula-agent, the default location of the
    PoP signing key. ProtectSystem=strict made it read-only, so every
    re-enrollment failed its preflight write test. Ordinary polling only
    reads the existing key, which is why the unit shipped this way since
    /etc/nebula-agent was introduced without anyone noticing.
  - StartLimitIntervalSec=0, so repeated failures can never leave the unit
    stopped for good. RestartSec already bounds the retry rate.

juev (changes requested):
I found two merge-blocking issues in the rekey lifecycle.

1. **A concurrent host edit can be lost when enrollment completes.**

`enrollHostInTx` unconditionally clears `pending_rekey`. The following sequence leaves the host on a stale certificate:

- Enrollment reads groups A and signs the certificate.
- A concurrent PATCH stores groups B and sets `pending_rekey=true`.
- The earlier enrollment commits the certificate containing groups A and clears `pending_rekey`.

The database now contains groups B, but the agent holds a potentially more permissive certificate containing groups A, with no further rekey scheduled.

Please bind the enrollment token to a host identity revision, or re-check the certificate-bound fields inside the enrollment transaction and preserve `pending_rekey` when they changed after signing.

2. **Failed re-enrollment retries without any backoff.**

After `Reenroll` returns an error, `startPoller` immediately continues and creates another `Poller`. `Poller.Run` performs its first poll immediately, so neither `PollInterval` nor systemd’s `RestartSec` limits these retries.

A persistent local error such as an unwritable target directory therefore creates a tight loop:

`poll → mint token → preflight failure → immediate poll`

This produces an unbounded stream of authenticated requests, SQLite nonce/token writes, and log entries. Please add a context-aware, bounded backoff before retrying the rekey.

**Code pointers**

- `internal/store/sqlite.go:1913` — unconditional `pending_rekey=0`
- `internal/api/enroll.go:93-215` — host identity is read and signed before the enrollment transaction
- `cmd/nebula-agent/main.go:387-399` — immediate retry after `Reenroll` failure
- `internal/agent/poller.go:212-228` — every new poller polls immediately

juev (approved):

juev:
thx!

PR #324, opened 2026-07-28
fix(cliargs): add guards for unknown commands and stray positional ar…

inode64:
I was performing an enrollment on an Ubuntu client and accidentally typed "enrool" instead of "enroll"; this caused the CLI to hang without displaying any error, and I drove myself crazy trying to figure out what was going on.

## Checklist

- [x] New behaviour has tests

### DIFF
--- a/cmd/nebula-agent/main.go
+++ b/cmd/nebula-agent/main.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/forgekeep/nebula-mesh/internal/agent"
+	"github.com/forgekeep/nebula-mesh/internal/cliargs"
 	"github.com/forgekeep/nebula-mesh/internal/config"
 	"github.com/forgekeep/nebula-mesh/internal/models"
 	"github.com/forgekeep/nebula-mesh/internal/version"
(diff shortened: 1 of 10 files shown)


juev (changes requested):
The argument validation fix is useful, but the current redaction is incomplete and introduces a credential disclosure regression.

`Guard.RejectPositional` and `Guard.UnknownCommand` include the raw argument in the returned error unless it has an `nme_` or `nmi_` prefix. This can expose legacy enrollment tokens, passwords, API keys, and backup passphrases in stderr or service logs.

For example:

```text
go run ./cmd/nebula-mgmt user create TEST_PASSWORD_DO_NOT_USE --api-key test-key
error: unexpected argument "TEST_PASSWORD_DO_NOT_USE" after flags; ...
```

Legacy enrollment tokens are also still accepted by the server, but `bootstraptoken.PurposeOf` intentionally cannot recognize them, so the following error prints the token:

```text
nebula-agent <legacy-token> --server https://mgmt.example.com
error: unknown command "<legacy-token>"; ...
```

Please make usage errors omit the raw unknown/positional value. Since arbitrary passwords and legacy tokens have no reliable prefix, trying to identify every secret format is not sufficient.

Please also add negative regression tests covering:

- an unprefixed legacy enrollment token;
- a positional password or API key passed to `nebula-mgmt`;
- confirmation that the credential is absent from both the returned error and command output.

There is also an invariant scope mismatch: `SEC-SECRET-001` explicitly excludes authentication tokens and login passwords, while the new token-redaction tests are added as its anchors. Please either introduce a dedicated credential-disclosure invariant or update the existing invariant’s rule and scope so the normative text matches the tests.

Code pointers:

- `internal/cliargs/cliargs.go`: `RejectPositional`, `UnknownCommand`, `Redact`
- `internal/bootstraptoken/token.go`: `PurposeOf`, legacy enrollment handling
- `internal/cliargs/cliargs_test.go`: `TestRedactsBootstrapTokens`
- `docs/security/invariants.md`: `SEC-SECRET-001`

juev (approved):
Some of the issues existed before your PR.
But your PR made them more critical. I accept your changes, but will make some adjustments to address the issues.
Thank you!

PR #342, opened 2026-07-29
feat(firewall): add cidr and local_cidr fields to firewall rules for enhanced traffic scoping

inode64:
With the latest changes since Nebula 1.10.0 (default_local_cidr_any now defaults to false, meaning that any firewall rule intended to target an unsafe_routes entry must explicitly declare it via the local_cidr field), it is now necessary to define `local_cidr` or `cidr` rules to enable routing between nodes.

## Checklist

- [x] New behaviour has tests


juev:
thx!

PR #348, opened 2026-07-30
feat(pki): sign unsafe networks into host certificates

inode64:
Nebula authorizes routing on the certificate, not on config. A gateway whose cert omits a prefix silently refuses to route it, and both ends drop the traffic in Firewall.Drop as ErrInvalidLocalIP -- before any firewall rule is evaluated, because routableNetworks is built from the cert's networks plus its unsafe networks.

The signer never populated TBSCertificate.UnsafeNetworks and the Host model had no field to express it, so advanced.unsafe_routes (the consumer half) could never work end to end: the route was installed on the client, the local_cidr firewall rule was rendered on the gateway, and every packet was still dropped with no log line -- a failure that reads like a peer or LAN outage rather than a configuration gap.

Add Host.UnsafeNetworks as certificate-bound state alongside Groups and NebulaIPs:

- pki.SignRequest carries UnsafeNetworks into the signed body, wired through all three signing paths (enroll, renewal, mobile bundle) via a shared models.ParseUnsafeNetworks, so an agent and a mobile bundle canonicalize the same stored value identically.
- Migration 028 adds hosts.unsafe_networks_json, defaulting to '[]'. Existing certs carry none, so no cert becomes stale.
- CertificateIdentity gains the field, so editing it schedules a re-issuance exactly like editing groups or addresses. Both mesh-import write paths persist it too -- they are hand-written column lists separate from CreateHost/UpdateHost, and without this an imported gateway would lose its routing authority at the next re-issuance.
- models.ValidateUnsafeNetworks takes the network's overlay CIDRs and checks both the prefixes and their overlap with the overlay in one call, so no write path can run half the validation. It rejects host bits, duplicates, self-overlap and more than 16 entries. Canonical form is required rather than silently masked: an operator typing 192.168.1.1/24 means the gateway's address, not the prefix it serves.
- The web form field reuses a shared splitCommaList that also replaces both inline copies of the groups splitter. It separates on commas and line breaks only -- never spaces, which would break a group name like "my group" into two on every save.

docs/agent.md documents all three pieces an unsafe route needs and on which host each one lives; the firewall section previously described the local_cidr rule as sufficient when it is only necessary. It also spells out that certificate-bound edits set pending_rekey and make the agent re-enrol with a fresh keypair (ADR 0004 §7.1) rather than being a server-side re-sign -- the agent has to be running and reachable for the change to take effect.

Verified end to end against the production mesh: with the prefix absent the client routes and the gateway's local_cidr rules are both still in place and traffic is dropped; adding it to the certificate is the only change that restores reachability.

## Checklist

- [x] New behaviour has tests


juev:
thx!

PR #369, opened 2026-08-31
feat(config): add punchy.respond configuration option for NAT traversal

inode64:
## Checklist

- [x] New behaviour has tests


juev:
Thanks @inode64 — nicely done. The field lands through every layer following the existing tri-state pattern (model, generator, web UI, audit diff, import/reconcile), and the tests pin the unset case so existing configs render unchanged.

Merging now. We'll follow up with a small docs pass (openapi spec, README, agent docs) in a separate PR.

PR #372, opened 2026-09-01
feat(windows): add graphical installer and PowerShell script for Nebula Mesh

inode64:
…
Tested on Windows 10, for now.

## Checklist

- [x] New behaviour has tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,14 +82,18 @@ jobs:
   windows-installer:
     needs: goreleaser
     runs-on: windows-latest
+    # Same approval gate as goreleaser: this job also writes to the release.
+    environment: release
     permissions:
       contents: write
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
 
+      # Pinned so a release is reproducible and so CI compiles the .iss with
+      # the same compiler the release does. Bump it here and in ci.yml together.
       - name: Install Inno Setup
         shell: pwsh
-        run: choco install innosetup --no-progress --yes
+        run: choco install innosetup --version 6.7.1 --no-progress --yes
 
       # -AgentVersion is pinned to the tag being released so the installer a
       # user downloads from a release deploys that release's agent, not
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,3 +75,48 @@ jobs:
           args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  # The graphical installer is not a Go binary, so GoReleaser cannot produce
+  # it: it is an Inno Setup compile that only runs on Windows. Built after the
+  # release exists, then attached to it.
+  windows-installer:
+    needs: goreleaser
+    runs-on: windows-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Install Inno Setup
+        shell: pwsh
+        run: choco install innosetup --no-progress --yes
+
+      # -AgentVersion is pinned to the tag being released so the installer a
+      # user downloads from a release deploys that release's agent, not
+      # whatever is newest whenever they happen to run it. Nebula stays on
+      # "latest": it is a separate project on its own cadence.
+      - name: Build the installer
+        shell: pwsh
+        env:
+          TAG: ${{ github.ref_name }}
+        run: |
+          $version = $env:TAG -replace '^v', ''
+          .\deploy\windows\build-installer.ps1 -Version $version -AgentVersion $env:TAG
+          $exe = "dist\windows-installer\nebula-mesh-setup-$version.exe"
+          if (-not (Test-Path -LiteralPath $exe)) { throw "installer not produced: $exe" }
+          # sha256sum's own format, so `sha256sum -c` verifies the download as-is.
+          $hash = (Get-FileHash -Algorithm SHA256 -LiteralPath $exe).Hash.ToLowerInvariant()
+          Set-Content -Path "$exe.sha256" -Encoding ascii `
+            -Value "$hash  nebula-mesh-setup-$version.exe"
+          "INSTALLER=$exe" >> $env:GITHUB_ENV
+
+      - name: Attach to the release
+        shell: pwsh
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ github.ref_name }}
+        run: |
+          # Via a local: "$env:INSTALLER.sha256" would parse the dot as part of
+          # the environment variable's name and expand to nothing.
+          $exe = $env:INSTALLER
+          gh release upload $env:TAG $exe "$exe.sha256" --clobber
--- a/deploy/windows/Install-NebulaMesh.ps1
+++ b/deploy/windows/Install-NebulaMesh.ps1
@@ -1,0 +1,1245 @@
+<#
+.SYNOPSIS
+    Downloads and installs Nebula and nebula-agent on Windows, enrolls the host
+    against a management server, and registers both as Windows services.
+
+.DESCRIPTION
+    One-shot installer for a Windows host that should join a Nebula Mesh network:
+
+      1. downloads nebula-agent (forgekeep/nebula-mesh) and Nebula
+         (slackhq/nebula) release archives and verifies their SHA-256 checksums;
+      2. installs the binaries below %ProgramFiles% and hardens the ACLs so only
+         SYSTEM and local Administrators can write them (the services run as
+         LocalSystem, so a user-writable binary would be a local privilege
+         escalation path);
+      3. prompts for the management server URL and the one-time enrollment
+         token, hands the token to `nebula-agent enroll` through an owner-only
+         file, and deletes that file afterwards;
+      4. points agent.yml's reload hook at the Nebula service (Windows has no
+         SIGHUP, so nebula_pid_file cannot be used);
+      5. registers and starts the NebulaMeshAgent and nebula services.
+
+    Re-running the installer over an existing installation upgrades the binaries
+    and re-registers the services; enrollment is skipped unless -Force is given.
+
+.EXAMPLE
+    .\Install-NebulaMesh.ps1
+
+    Fully interactive: asks for the server URL, the token, and which services to
+    enable.
+
+.EXAMPLE
+    .\Install-NebulaMesh.ps1 -ServerUrl https://mgmt.example.com:8080 -TokenFile C:\provision\enroll.token -Unattended
+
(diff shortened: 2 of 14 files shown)


deploy/windows/Install-NebulaMesh.ps1
```diff
@@ -0,0 +1,1245 @@
...
+# agent.yml is a flat YAML map, so rewriting these two keys in place is safe.
+function Set-AgentReloadCommand([string]$AgentConfigPath, [string]$NebulaService) {
+    if (-not (Test-Path -LiteralPath $AgentConfigPath)) {
+        Write-Note ("no agent.yml yet - after enrolling, set nebula_reload_command so config updates " +
+            'reach Nebula (see deploy/windows/README.md).')
+        return
+    }
+
+    $net = '"%SystemRoot%\System32\net.exe"'
+    $command = "$net stop $NebulaService & $net start $NebulaService"
```
juev: **Blocker.** This is analysis, not something I reproduced on Windows — please verify before acting on it.

The generated value is:

```
"%SystemRoot%\System32\net.exe" stop nebula & "%SystemRoot%\System32\net.exe" start nebula
```

and the agent runs the hook as `cmd /C ` + that string verbatim (`internal/agent/reload_windows.go`, via `SysProcAttr.CmdLine`). Per `cmd /?`, quotes are preserved only when there are *exactly two* quote characters with an executable name between them. There are four here, so the other documented rule applies: cmd strips the leading quote and the last quote on the line, leaving

```
%SystemRoot%\System32\net.exe" stop nebula & "%SystemRoot%\System32\net.exe start nebula
```

The `&` then sits inside an unterminated quote and neither half is a valid command. If that reading is right, the reload hook fails on every managed config change, the poller never acks the config, and the installer's central promise — config updates reaching Nebula — quietly does not hold.

`%SystemRoot%\System32\net.exe` contains no spaces, so the simplest fix is to drop the quotes entirely.

Worth pinning with a test: `reload_script_windows_test.go` does not cover this shape because `scriptTouch` starts with `type`, not with a quote. The existing "Windows agent service" job already runs `TestNebulaReloader` on a real Windows host, so a case whose command starts with a quoted executable path would catch it for good.

deploy/windows/README.md
```diff
@@ -0,0 +1,238 @@
...
+takeown /F "C:\Program Files\Nebula Mesh" /R /A
+icacls "C:\Program Files\Nebula Mesh" /reset /T /C
+```
+
+Because Windows has no `SIGHUP`, the installer rewrites `agent.yml` to reload
+Nebula through its service instead of a PID file:
+
+```yaml
+nebula_pid_file: ""
+nebula_reload_command: '"%SystemRoot%\System32\net.exe" stop nebula & "%SystemRoot%\System32\net.exe" start nebula'
```
juev: Same command as in `Set-AgentReloadCommand` — see the blocker comment on `Install-NebulaMesh.ps1`. A `cmd /C` command line that starts with a quote and contains four quotes gets mangled by cmd's quote-stripping rule, so this documented snippet would not work if copied by hand either. Please fix both together.

deploy/windows/Install-NebulaMesh.ps1
```diff
@@ -0,0 +1,1245 @@
...
+                if ($web.Response) { $code = [int]$web.Response.StatusCode }
+                Set-Failure 'server-not-mgmt' @(
+                    "$probe answered HTTP $code.",
+                    'A Nebula Mesh management server answers 200 there. Something else is',
+                    'listening on that address, or the URL needs a different port or path.')
+            }
+            default {
+                Set-Failure 'server-unreachable' @(
+                    "Cannot reach $probe.",
+                    $_.Exception.Message)
```
juev: `$_` inside `switch ($status)` is bound to the switch input — the status string — not to the `ErrorRecord` from the enclosing `catch`. With `Set-StrictMode -Version 3.0` in force, `$status.Exception` raises a PropertyNotFound error, so this branch replaces the intended "Cannot reach ..." message with a strict-mode failure.

This is the branch that handles every unclassified network failure, including the common case where `Get-WebException` returns `$null` and `$status` is `''`. Capturing the record before the switch fixes it:

```powershell
} catch {
    $record = $_
    ...
    default {
        Set-Failure 'server-unreachable' @(
            "Cannot reach $probe.",
            $record.Exception.Message)
    }
```

internal/models/mesh_import.go
```diff
@@ -47,6 +47,69 @@ func (p AgentProfile) Validate() error {
...
 }
 
+// isCleanAbsoluteAgentPath reports whether value is an absolute, already-clean
+// path in either POSIX or Windows form.
+//
+// path/filepath is deliberately NOT used here. It compiles to the *server's*
+// GOOS, while these paths describe the *agent's* filesystem — so a Windows
+// agent enrolling against a Linux server was rejected with "invalid agent
+// profile" for C:\ProgramData\Nebula\config.yml, which filepath.IsAbs on Linux
+// cannot recognise. The mirror case fails the same way, which is why the
```
juev: `recognise` -> `recognize`. The `misspell` linter fails on it, which reds the `static / security` job and, through it, `build / test / lint`. Two more in `agent_profile_test.go` (lines 89 and 102, `flavour`).

internal/models/mesh_import.go
```diff
@@ -47,6 +47,69 @@ func (p AgentProfile) Validate() error {
...
+		// filepath.Clean. A bare root names a directory, never one of the
+		// four files the profile points at.
+		return value != "/" && path.Clean(value) == value
+	}
+	return isCleanAbsoluteWindowsPath(value)
+}
+
+// isCleanAbsoluteWindowsPath accepts a drive-rooted path (C:\dir\file) or a UNC
+// share path (\\server\share\file).
+func isCleanAbsoluteWindowsPath(value string) bool {
```
juev, line 93: Non-blocking, but worth a thought while the rule is being widened.

`Validate` enforces "the four paths must differ" by comparing strings. That holds for POSIX paths, but NTFS is case-insensitive and drops trailing dots and spaces, so `C:\ProgramData\Nebula\host.key` and `C:\ProgramData\Nebula\HOST.KEY` now validate as two distinct paths while naming one file. `\\?\C:\...` and `\\.\...` device paths also pass this check.

Practical impact is small — an agent can only hurt itself this way — but if the distinctness rule is worth having, folding case for the Windows form before the uniqueness comparison would keep it honest.

internal/models/agent_profile_test.go
```diff
@@ -21,3 +24,91 @@ func TestAgentProfileDefaultsAndValidation(t *testing.T) {
...
+			NebulaCertPath:   `C:\ProgramData\Nebula\host.crt`,
+			NebulaKeyPath:    `C:\ProgramData\Nebula\host.key`,
+		}
+		if err := profile.Validate(); err == nil {
+			t.Errorf("%s: accepted %q", name, bad)
+		}
+	}
+}
+
+// The four paths must still be distinct, in either flavour.
```
juev: `flavour` -> `flavor` here and on line 102 — `misspell` is what is failing CI.

deploy/windows/nebula-mesh-setup.iss
```diff
@@ -0,0 +1,738 @@
...
+end;
+
+{ Writes the token where only SYSTEM and Administrators can read it, so the
+  secret never sits in a command line or in a world-readable temp file. }
+function WriteTokenFile: String;
+var
+  Path: String;
+  ResultCode: Integer;
+begin
+  Path := ExpandConstant('{tmp}\enroll.token');
```
juev, line 584: The token is written first and the ACL applied afterwards — the opposite order from `Invoke-AgentEnroll`, which creates the file empty, calls `Protect-File`, and only then writes the secret. `{tmp}` already lives under the elevated administrator's profile, so the exposure window is narrow, but matching the script costs nothing and keeps one rule instead of two.

`ResultCode` from the `icacls` call is also ignored, so a failed hardening passes silently.

.github/workflows/release.yml
```diff
@@ -75,3 +75,48 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  # The graphical installer is not a Go binary, so GoReleaser cannot produce
+  # it: it is an Inno Setup compile that only runs on Windows. Built after the
+  # release exists, then attached to it.
+  windows-installer:
+    needs: goreleaser
+    runs-on: windows-latest
+    permissions:
```
juev, line 87: The `goreleaser` job runs under `environment: release`, which is where the manual approval gate for this repository lives. This job writes release assets with `contents: write` outside that environment. It does `needs: goreleaser`, so the approval has effectively happened by the time it runs, but a release artifact ends up produced outside the protected environment. Adding `environment: release` here would keep the gate uniform.

Separately, `choco install innosetup` is unpinned on a release path — the compiler version that builds a published setup.exe is whatever Chocolatey serves that day.

juev (changes requested):
Read the whole diff plus the two failing checks.

Overall this is careful work. The ACL handling (the empty-DACL trap on leaf files, deliberately not using `icacls /T`), the order in which the token file is created and hardened, the `/healthz` pre-flight before the single-use token is spent, and the `!!` failure-code protocol between the script and the wizard are all things you only get right after using Windows in anger. I checked the release asset names and checksum-file formats against the current releases of both projects, and the wintun layout inside `nebula-windows-amd64.zip` — those all match what the script expects.

Two things block a merge; a few smaller ones are worth a look. Details are in the inline comments.

One request for the description: the `AgentProfile.Validate()` change is a separate and genuinely valuable bug fix — a Windows agent enrolling against a Linux management server really was rejected with HTTP 400 because `filepath.IsAbs` compiles to the server's GOOS. Under a "Windows installer" title that is easy to miss. Please call it out explicitly, or split it into its own PR.

deploy/windows/nebula-mesh-setup.iss
```diff
@@ -0,0 +1,751 @@
...
+    end;
+
+    SaveInstallLog;
+
+    { Not RaiseException: that prefixes the text with "Runtime error (at x:y)",
+      which buries the explanation under something the operator cannot act on.
+      Show the diagnosis on its own, then carry the failure to the final page.
+      The files are installed either way - only the enrollment step failed, and
+      it can be retried without reinstalling. }
+    if ResultCode <> 0 then
```
juev, line 698: **Blocker.** In an unattended installation, this branch converts a failed PowerShell run into a successful Setup exit. For example, with `/VERYSILENT`, an unreachable management server or failed enrollment makes the script return a non-zero `ResultCode`, but this code only records `InstallFailed` and shows a message box. `CurStepChanged` then returns normally, so Inno Setup completes with exit code 0.

Provisioning systems will mark the host as deployed even though the script may have stopped before downloading binaries or enrolling it. Please abort the Setup process with a non-zero exit when the script fails, while retaining the actionable error message for interactive installs.

inode64:
Sorry, I don't know what I pressed regarding the messages; check if the code looks right now.

juev (approved):
thx!

PR #376, opened 2026-09-08
fix(hosts): explain duplicate host names instead of returning a generic 500

inode64:
Creating a host whose name was already taken in its network failed with a bare "500 Failed to create host": the UNIQUE(network_id, name) violation reached the handlers as a raw SQLite error, leaving an operator unable to tell a typo from an outage. Renaming onto a taken name failed the same way.

Both paths now follow the two-layer shape the codebase already uses for overlay addresses:

- A friendly pre-check (conflictingHostName, mirroring validateHostIPs) resolves the collision before the write and names the incumbent, on create and on rename, in both the web form and the API. A host keeping its own name is excluded, so an edit that leaves the name alone still saves.

- The UNIQUE index remains the authority, since the pre-check can be raced. CreateHost, CreateHostAndToken and UpdateHost map the violation to ErrDuplicateEntry, which the transports answer with an inline form error (web) or 409 (API). UpdateHost mapped nothing at all before, which is why a colliding rename surfaced as a 500.

Attribution is confirmed rather than assumed: a unique violation triggers a lookup before it is blamed on the name. UNIQUE(network_id, name) is the table's only unique index today, but a later migration adding another would otherwise make this silently misreport an unrelated failure. Unconfirmed, the raw error stands and the caller reports a server fault. Non-constraint failures never reach the path at all: a full disk raises SQLITE_FULL, not SQLITE_CONSTRAINT_UNIQUE.

Adds GetHostByName to the store, and documents the 409 that POST /api/v1/hosts has been able to return since ErrIPTaken.

## Checklist

- [x] New behaviour has tests

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -324,6 +324,15 @@ paths:
                 $ref: "#/components/schemas/CreateHostResponse"
         "400":
           $ref: "#/components/responses/Error"
+        "409":
+          description: >-
+            The name is already taken by another host in the network, one of
+            the requested nebula_ips is already assigned, or a mesh import is
+            collecting for this network.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
 
   /api/v1/hosts/{id}:
     get:
--- a/internal/api/host_duplicate_name_test.go
+++ b/internal/api/host_duplicate_name_test.go
@@ -1,0 +1,125 @@
+package api
+
+import (
+	"bytes"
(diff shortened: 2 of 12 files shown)


juev (approved):
Thank you, @inode64, for fixing this and adding regression tests for both creation and renaming. I reproduced the previous 500 responses on main, and all store, API, and web tests passed with the race detector on this PR. Merging this for v0.16.2.

